PR #8, opened 2025-02-05
ctx/add union monitoring

ctxswitch:
* [x] Enable several prometheus rules
* [x] Add support for both Loki and Fluentbit.  These will be enabled by default.
* [x] Adds a the Loki datasource to grafana.
* [x] Fixes the spelling of the operator container.  I noticed that it was misspelled as we are now using the container name as the service name in the logs.
* [ ] Put together some union service dashboards.
* [ ] Clean up the metrics dashboards.  I may revisit having the Kube monitoring stack create them and statically define the ones we want in the charts.

## Logs
Screenshot 2025-02-04 at 4 04 35 PM

### DIFF
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -421,7 +421,7 @@ fluent-bit:
           Name loki
           Match kube.*
           Host unionai-dataplane-loki.union.svc.cluster.local
-          Labels service_name=$kubernetes['container_name']
+          Labels service_name=$kubernetes['container_name'] namespace=$kubernetes['namespace_name']
 
       [OUTPUT]
           Name loki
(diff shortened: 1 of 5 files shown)


charts/dataplane/values.yaml
```diff
@@ -365,6 +365,69 @@ dcgmExporter:
...
+    enabled: false
+  dashboards:
+    enabled: true
+    labelKey: grafana_dashboard
+    labelValue: 1
+    annotations: { }
+    namespace: union
+    deterministicUid: false
+  config:
+    service: |
```
mhotan, line 380: Non-blocking Q: We seem to be duplicating this block from the fluent-bit [defaults](https://github.com/fluent/helm-charts/blob/8440ebf9dae40285864ba697939d515648a7aaca/charts/fluent-bit/values.yaml#L377-L430). Is that a correct observation? If so, is that intentional?

mhotan (approved):

charts/dataplane/values.yaml
ctxswitch: That was intentional.  I'm leaning into being explicit with the configurations and I do actually expect this to potentially change as the charts mature.

PR #38, opened 2025-03-06
ctx/feat(dataplane): add datacatalog service

ctxswitch:
* [x] Adds support for the data catalog service in the dataplane.
* [x] Add a sandbox postgres service to sandbox for testing.
* [x] bump dataplane to 2025.2.11
* [x] bump sandbox to 2025.2.1


mhotan (approved):

katrogan:
sorry for the drive by review, but in closed source we use cache service instead of datacatalog but both components run in the control plane. I'm curious why we choose to deploy datacatalog in the dataplane?

PR #44, opened 2025-03-12
chore(aws-example): switch to api-based eks auth

ctso:
I was running into problems using this example with ConfigMap-based authentication.  Instead, let's just use the newer API-based method.


providers/aws/variables.tf
```diff
@@ -36,11 +36,6 @@ variable "admin_role_arns" {
   default = []
 }
 
-variable "admin_user_arns" {
-  type    = list(string)
-  default = []
-}
-
```
ctso, line 43: Figure most organizations are not using IAM users, so dropping support for this for now.

providers/aws/storage.tf
```diff
@@ -1,9 +1,13 @@
...
+  byte_length = 4
+}
+
 module "union-data" {
   source  = "terraform-aws-modules/s3-bucket/aws"
   version = "4.6.0"
 
   # TODO: change me to something different (var)
-  bucket                  = "${local.name_prefix}-data"
+  bucket                  = "${local.name_prefix}-data-${random_id.s3_bucket.hex}"
```
ctso, line 10: Ran into some issues here where the bucket already exists, let's just suffix this with a random string for now.

providers/aws/eks.tf
```diff
@@ -14,6 +41,9 @@ module "eks" {
   vpc_id     = module.vpc.vpc_id
   subnet_ids = module.vpc.private_subnets
 
+  authentication_mode = "API"
```
mhotan, line 44: 🆒

providers/aws/provider.tf
```diff
@@ -28,21 +28,3 @@ provider "helm" {
     }
   }
 }
-
-provider "kubernetes" {
-  host                   = module.eks.cluster_endpoint
-  cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
-  token                  = data.aws_eks_cluster_auth.default.token
-  # exec {
```
mhotan, line 36: nit: Can we remove these commented out providers now?

mhotan (approved):

providers/aws/provider.tf
ctso: This whole block is already being removed, commented out stuff is going away with this change.

PR #47, opened 2025-03-13
fix(helm): support vendor-specific versions of k8s

ctso:
When installing these charts on an AWS EKS cluster, we're currently failing with

```
Error: chart requires kubeVersion: >= 1.28.0 which is incompatible with Kubernetes v1.32.2-eks-bc803b4
```

Changing the constraint to `>= 1.28.0-0` allows us to support pre-release and vendor-specific versions.

### Test Plan
Tested this locally, the fix works.

### Internal Use
fixes D-1111

### DIFF
--- a/charts/dataplane-crds/Chart.yaml
+++ b/charts/dataplane-crds/Chart.yaml
@@ -5,4 +5,4 @@ type: application
 icon: "https://i.ibb.co/JxfDQsL/Union-Symbol-yellow-2.png"
 version: 2025.1.2
 appVersion: 2025.1.2
-kubeVersion: ">= 1.28.0"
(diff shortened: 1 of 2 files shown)


mhotan (approved):

davidmirror-ops:
This works but still logs the error on installation:
```shell
Release "unionai-dataplane-crds" does not exist. Installing it now.
Error: chart requires kubeVersion: >= 1.28.0 which is incompatible with Kubernetes v1.32.2-eks-bc803b4

NAME: unionai-dataplane
LAST DEPLOYED: Mon Mar 17 16:13:18 2025
NAMESPACE: union
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

PR #49, opened 2025-03-13
set route prefix to /prometheus

troychiu:
set route prefix to /prometheus


charts/dataplane/values.yaml
```diff
@@ -848,6 +848,7 @@ prometheus:
         requests:
           cpu: "4"
           memory: "8Gi"
+      routePrefix: /prometheus
```
mhotan, line 851: I
```suggestion
      routePrefix: /prometheus/
```

mhotan (approved):
It be g

PR #52, opened 2025-03-14
Dcgm exporter service monitor

troychiu:
add Dcgm exporter service monitor


charts/dataplane/templates/_helpers.tpl
```diff
@@ -522,11 +522,11 @@ access the storage is injected.
 {{- end -}}
 
 {{- define "prometheus.health.url" -}}
-http://union-prometheus:9090/-/healthy
+http://{{ include "union-operator.fullname" . }}-prometheus:80/-/healthy
```
davidmirror-ops, line 525: @troychiu are these the fixes for TLM to work?

PR #54, opened 2025-03-14
chore(charts/dataplane): add dependency on k8s-sigs metrics-server chart

ctso:
We depend on `metrics-server` for billable usage.  Because `metrics-server` creates an `APIService`, we can only run this component once per cluster.  This change assumes cluster administrators will likely already be running `metrics-server`, so we won't install it by default.

Should cluster administrators not want to worry about installing `metrics-server` themselves, they can set:
```
monitoring:
  metricsServer:
    enabled: true
```

And this chart will install it.

### Test Plan
Tested this against `byok-aws`

### Internal Use
fixes D-1124

### DIFF
--- a/charts/dataplane/Chart.yaml
+++ b/charts/dataplane/Chart.yaml
@@ -14,5 +14,5 @@ dependencies:
   - name: metrics-server
     repository: https://kubernetes-sigs.github.io/metrics-server/
     version: 3.12.2
-    condition: monitoring.metricsServer.enabled
-    alias: metricsServer
+    condition: metrics-server.enabled
+    alias: metrics-server
--- a/charts/dataplane/Chart.yaml
+++ b/charts/dataplane/Chart.yaml
@@ -12,3 +12,8 @@ dependencies:
     version: 68.2.2
     condition: monitoring.prometheus.enabled
     alias: prometheus
+  - name: metrics-server
+    repository: https://kubernetes-sigs.github.io/metrics-server/
+    version: 3.12.2
+    condition: monitoring.metricsServer.enabled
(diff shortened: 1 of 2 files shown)


charts/dataplane/values.yaml
```diff
@@ -712,6 +712,13 @@ loki:
   #   replicas: 0
   # isDefault: true
 
+monitoring:
+  prometheus:
+    enabled: true
```
ctso, line 717: We're referencing this condition in `Chart.yaml` but the value didn't exist.  It still seemed to be installing `kube-prometheus-stack`, but might as well define this to be safe.

charts/dataplane/Chart.yaml
```diff
@@ -10,5 +10,9 @@ dependencies:
   - name: kube-prometheus-stack
     repository: https://prometheus-community.github.io/helm-charts
     version: 68.2.2
-    condition: monitoring.prometheus.enabled
```
ctso, line 13: We were not setting this, and the condition was causing us to install `kube-prometheus-stack` even if it wasn't set (`null`).  For now, I think we want to assume we'll always install this chart as a dependency

charts/dataplane/values.yaml
```diff
@@ -712,6 +712,10 @@ loki:
   #   replicas: 0
   # isDefault: true
 
+monitoring:
```
ctxswitch, line 715: This is aliased as metricsServer so you don't need to nest this.  I think we can just add enabled to metricsServer and keep it at the root with all of the other metrics server settings instead of adding an additional tree.

ctxswitch (approved):

PR #67, opened 2025-03-17
ctx/chore(release): version bump, commit, and merge.

ctxswitch:
* [x] The beginning of the tooling to manage our releases. Relatively simplistic to start out, but should be enough of a base to let us start building out a release pipeline in buildkite.
* [ ] Update actions to only run when branches named `release/*` are pushed into main.


tasks/__init__.py
```diff
@@ -0,0 +1,5 @@
+from invoke import Collection
+import tasks.builder.base as base
+
+ns = Collection()
+ns.add_collection(Collection.from_module(base), name="builder")
```
mhotan, line 5: nice

mhotan (approved):

PR #72, opened 2025-03-18
ctx/chore: add pre-commit hooks and format/lint

ctxswitch:
* [x] Adds pre-commit hooks to run formatters and other linters in the future.  I may consider bundling helm-docs into this as well instead of generating those in actions.


.pre-commit-config.yaml
pmahindrakar-oss, line 1: Can we use the existing make commands inside this precommit file like format_check_python_code which already does some of these checks. That would be easier to maintain as the same checks would also be running in buildkite

pmahindrakar-oss: oh .this is another repo and not the cloud repo. Still might useful if we can leverage the same tooling here

pmahindrakar-oss (approved):

.pre-commit-config.yaml
ctxswitch: I'll look at that and update later - I have one more update to get out real quick.  For what it's worth, these will most likely never be running through buildkite - even in the release pipelines.

PR #78, opened 2025-03-24
enable k8s secret as union secret backend by default

troychiu:
enable k8s secret as union secret backend by default


charts/dataplane/templates/operator/namespaces.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
```
ctso, line 2: Was there a specific issue you ran into where we need to add a `Namespace` to the Helm chart here?  Typically helm charts will leave it up to the user to ensure the namespace exists.

charts/dataplane/templates/operator/serviceaccount-proxy-secret.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if and .Values.config.proxy.smConfig.enabled (eq .Values.config.proxy.smConfig.type "K8s") }}
```
ctxswitch, line 1: I missed this the first time through.  I'm wondering if we should move this enabled check into `proxy` and get both values from there.  i.e.:

instead of

```yaml
proxy:
  secretsNamespace: ""
```

use

```yaml
proxy:
  secretManager:
    enabled: true
    namespace: ""
```

charts/dataplane/values.yaml
```diff
@@ -326,6 +331,13 @@ config:
         endpoint: '{{ include "proxy.health.url" . }}'
       prometheus:
         endpoint: '{{ include "prometheus.health.url" . }}'
+  proxy:
+    smConfig:
+      enabled: true
```
ctxswitch, line 336: Based on the previous comment this can be set to the enabled value in the proxy block

PR #82, opened 2025-03-26
ctx/fix(dataplane): remove newline trim when using fastRegistrationURL

ctxswitch:
* [x] Removes an inadvertent trim from the `fastRegistrationURL` block that caused the rendered config value to be added to previous line.


ctso (approved):
ugh, helm

PR #87, opened 2025-03-28
fix(serving): add permissions for serving, fix configuration

ctso:
- Adds RBAC permissions to interact with KnativeServing resources when serving is enabled
- Prefers to use top-level `serving.enabled` flag for a better UX

### DIFF
--- a/charts/dataplane/templates/operator/configmap.yaml
+++ b/charts/dataplane/templates/operator/configmap.yaml
@@ -21,10 +21,6 @@ data:
     authorizer: {{ tpl (toYaml .) $ | nindent 6 }}
   {{- end }}
     operator:
-      {{- if .Values.serving.enabled }}
-      apps:
-        enabled: true
-      {{- end }}
       enabled: {{ .Values.config.operator.enabled }}
       enableTunnelService: {{ .Values.config.operator.enableTunnelService }}
       tunnel:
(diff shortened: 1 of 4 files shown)


charts/dataplane/templates/operator/configmap.yaml
```diff
@@ -21,10 +21,6 @@ data:
     authorizer: {{ tpl (toYaml .) $ | nindent 6 }}
   {{- end }}
     operator:
-      {{- if .Values.serving.enabled }}
-      apps:
-        enabled: true
-      {{- end }}
       enabled: {{ .Values.config.operator.enabled }}
       enableTunnelService: {{ .Values.config.operator.enableTunnelService }}
       tunnel:
```
ctxswitch, line 26: This was from a previous PR, but should we move this back to the config values and handle it like apps?

ctxswitch (approved):

charts/dataplane/templates/operator/serviceaccount.yaml
```diff
@@ -86,6 +86,21 @@ rules:
       - /metrics
     verbs:
       - get
+  {{- if .Values.serving.enabled }}
```
EngHabu, line 89: How does it work today without this block?

ctso: It doesn't work, thus the fix here.  We are maintaining a separate Helm chart for BYOK, we have this block internally, but missed it in the BYOK chart that this repo maintains.

We do have plans to converge these charts.

PR #88, opened 2025-03-31
chore(serving): Move knative-operator to its own chart

ctso:
We want to avoid having the `dataplane` chart create lots of namespaces, and ideally we'd run `knative-operator` in a dedicated namespace.  This change introduces a new chart, which is just a wrapper around the `knative-operator` Kustomize manifests.

Users can opt to manage it themselves, or use our provided chart.  For now, we'll leave the CRDs bundled in `dataplane-crds` to avoid yet another CRD-specific chart.


ctxswitch:
We should probably pull the CRDs out.  If the user is managing the knative install themselves, including them in the dataplane crds may cause a conflict.

charts/knative-operator/templates/knative-operator.yaml
```diff
@@ -19,7 +40,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: operator-webhook-certs
-  namespace: {{ .Release.Namespace }}
+  namespace: knative-operator
```
ctxswitch, line 43: Suggestion: Non-blocking but we may want to add a `namespaceOverride` so the user can manage the location?

ctso:
> We should probably pull the CRDs out. If the user is managing the knative install themselves, including them in the dataplane crds may cause a conflict.


I thought about that, they are disabled by default though so I figured this would be easier than adding yet another `knative-operator-crds` chart.

charts/knative-operator/templates/knative-operator.yaml
ctso: 41c116ad6a0f30ecfd352dace6a7145c2aab1dfc

PR #99, opened 2025-04-21
ctx/feat(dataplane): add nodeobserver support

ctxswitch:
* [x] Add support for node observer service supporting accelerated datasets.


charts/dataplane/values.yaml
```diff
@@ -651,6 +651,56 @@ image:
...
+    requests:
+      cpu: "500m"
+      memory: "100Mi"
+  config:
+    criticalDaemonSets: []
+    #  - name: union-mountpoint
+    #    namespace: '{{ tpl .Release.Namespace . }}'
+    #  - name: unionai-dataplane-fluentbit
+    #    namespace: '{{ tpl .Release.Namespace . }}'
+
```
pmahindrakar-oss, line 703: to be removed?

pmahindrakar-oss (approved):
Can we use helm generated file if we want to check diffs for any changes to templates. We use it in our cloud repo already and helps immensely in testing and reviewing the impact of such template file changes

charts/dataplane/values.yaml
ctxswitch: No, comments to show usage.  It's common in other OSS charts to do this.

PR #105, opened 2025-05-02
ctx/fix(dataplane): allow prom to select anywhere

ctxswitch:
* [x] Open up monitor selectors to everything.  This fixes the issue with finding resources on externally managed/installed services.  Wide open for now, but will most likely have some sane defaults in the future.

### DIFF
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -769,6 +769,14 @@ prometheus:
     service:
       port: 80
     prometheusSpec:
+      ruleSelectorNilUsesHelmValues: false
+      ruleSelector: {}
+      scrapeConfigSelectorNilUsesHelmValues: false
+      scrapeConfigSelector: {}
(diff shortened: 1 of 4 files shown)


charts/dataplane/values.yaml
```diff
@@ -769,6 +769,14 @@ prometheus:
...
       port: 80
     prometheusSpec:
+      ruleSelectorNilUsesHelmValues: false
+      ruleSelector: {}
+      scrapeConfigSelectorNilUsesHelmValues: false
+      scrapeConfigSelector: {}
+      serviceMonitorSelectorNilUsesHelmValues: false
+      serviceMonitorSelector: {}
+      podMonitorSelectorNilUsesHelmValues: false
+      podMonitorSelector: {}
```
pmahindrakar-oss, line 779: May be we can add a comment with what you ran into here for reference

PR #108, opened 2025-05-06
ctx/chore(dataplane): set default prom retention to 3 days

ctxswitch:
* [x] Set the default prometheus retention to 3 days.


pmahindrakar-oss (approved):
Could you add reason we are reducing the retention.

PR #114, opened 2025-05-15
Union services control plane charts

pmahindrakar-oss:
* Adds green field templates working of the existing one in union cloud
* The templates are refactored to be able to install multiple services with single values file
* Adds helper to do recursive merges of the values
* Add override for controlplane chart to be installed with flag enabled for it
* Adds  generated tests

Tested this on internal EKS cluster and all services come up

### DIFF
--- a/charts/controlplane/templates/_helpers.tpl
+++ b/charts/controlplane/templates/_helpers.tpl
@@ -1,0 +1,381 @@
+{{- define "unionai.imagePullSecrets" -}}
+{{- if and (hasKey .config "imagePullSecrets") }}
+{{ toYaml .config.imagePullSecrets }}
+{{- else if and (hasKey .Values.global "imagePullSecrets") }}
+{{ toYaml .Values.global.imagePullSecrets }}
+{{- else }}
+{}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.resources" -}}
+{{- if and (hasKey .config "resources") }}
+{{ toYaml .config.resources }}
+{{- else if and (hasKey .Values.global "resources") }}
+{{ toYaml .Values.global.resources }}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.nodeSelector" -}}
+{{- if and (hasKey .config "nodeSelector") }}
+{{ toYaml .config.nodeSelector }}
+{{- else if and (hasKey .Values.global "nodeSelector") }}
+{{ toYaml .Values.global.nodeSelector }}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.service" -}}
+{{- if and (hasKey .config "service") }}
+{{ toYaml .config.service }}
+{{- else if and (hasKey .Values.global "service") }}
+{{ toYaml .Values.global.service }}
+{{- end }}
+{{- end }}
+
+
+{{- define "unionai.affinity" -}}
+{{- if and (hasKey .config "affinity") }}
+{{ toYaml .config.affinity }}
+{{- else if and (hasKey .Values.global "affinity") }}
+{{ toYaml .Values.global.affinity }}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.tolerations" -}}
+{{- if and (hasKey .config "tolerations") }}
+{{ toYaml .config.tolerations }}
+{{- else if and (hasKey .Values.global "tolerations") }}
+{{ toYaml .Values.global.tolerations }}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.env" -}}
+{{- if and (hasKey .config "env") -}}
+{{- toYaml .config.env -}}
+{{- else if and (hasKey .Values.global "env") }}
+{{- toYaml .Values.global.env -}}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.securityContext" -}}
+{{- if and (hasKey .config "securityContext") }}
+    {{- toYaml .config.securityContext -}}
+{{- else if and (hasKey .Values.global "securityContext") }}
+    {{- toYaml .Values.global.securityContext -}}
+{{- end }}
+{{- end }}
+
+{{- define "unionai.podSecurityContext" -}}
+{{- if and (hasKey .config "podSecurityContext") .config.podSecurityContext }}
+    {{- toYaml .config.podSecurityContext }}
+{{- else if and (hasKey .Values.global "podSecurityContext") .Values.global.podSecurityContext }}
+    {{- toYaml .Values.global.podSecurityContext }}
+{{- end }}
+{{- end }}
+
+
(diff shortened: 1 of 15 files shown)


charts/controlplane/templates/_helpers.tpl
```diff
@@ -0,0 +1,381 @@
...
+app.kubernetes.io/name: {{ include "unionai.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{- define "sync.fullname" -}}
+{{- default (printf "%s-sync" (include "unionai.name" .)) .Values.sync.fullnameOverride }}
+{{- end }}
+{{- define "sync.labels" -}}
+helm.sh/chart: {{ include "unionai.chart" . }}
+linkerd.io/inject: enabled
```
ctxswitch, line 236: Were we including linkerd?  I was under the impression that we were stripping out the requirement.

charts/controlplane/values.yaml
```diff
@@ -0,0 +1,308 @@
...
+      - artifacts
+      - serve
+      - --config
+      - /etc/config/*.yaml
+    configMap:
+      sharedService:
+        metrics:
+          scope: "artifacts:"
+      db:
+        host: "database-1-instance-1.c982kkq6ghd1.us-east-2.rds.amazonaws.com"
```
ctxswitch, line 130: Nit: we should consolidate these at the top level (maybe in a database block) and use a default value that isn't tied to something that we test on internally.

charts/controlplane/values.yaml
```diff
@@ -0,0 +1,308 @@
...
+      sampler:
+        parentSampler: traceid
+        traceIdRatio: 0.001
+    union:
+      internalConnectionConfig:
+        enabled: true
+        urlPattern: "{{ service }}.union.svc.cluster.local:80"
+      connection:
+        trustedIdentityClaims:
+          externalIdentityClaim: 0oa2n55xystvEjmOD5d7
```
ctxswitch, line 104: Nit: with this specific field, do we just leave it blank instead of having a value that has been transposed from an test system?

charts/controlplane/values.yaml
```diff
@@ -0,0 +1,308 @@
+# This section contains the default values for the controlplane chart.
+# Currently this includes only for union services and also values have been populated for testing only.
+global:
+  replicaCount: 1 # Keeping low replica count for testing purposes.
+  image:
+    repository: 479331373192.dkr.ecr.us-east-2.amazonaws.com/cloud # only for testing
```
ctxswitch, line 6: Nit: If this is for testing we probably shouldn't keep it here.  Probably want to just override this locally for your tests?

charts/controlplane/values.yaml
```diff
@@ -0,0 +1,308 @@
...
+        connectionPool:
+          maxIdleConnections: 20
+          maxOpenConnections: 20
+          maxConnectionLifetime: 1m
+      artifactsConfig:
+        app:
+          adminClient:
+            hackFlagUntilCellIsolation: true
+          artifactBlobStoreConfig:
+            container: artifacts-offloaded-byocp-test-us-east-2
```
ctxswitch, line 142: Nit: maybe something a little more generic?

charts/controlplane/values.yaml
```diff
@@ -0,0 +1,308 @@
...
+      - --config
+      - /etc/config/*.yaml
+    serviceAccount:
+      annotations:
+        eks.amazonaws.com/role-arn: arn:aws:iam::440744229819:role/staging-executions-role
+    configMap:
+      sharedService:
+        metrics:
+          scope: "executions:"
+      db:
```
ctxswitch, line 242: Nit: There's a lot of redundancy in the config maps that I think we could move out to a root level block.

charts/controlplane/templates/deployment.yaml
```diff
@@ -0,0 +1,172 @@
+{{- if (index .Values "controlplane" | default dict).enabled }}
```
ctxswitch, line 1: This is really slick.  I may update the dataplane to use something similar.

charts/controlplane/templates/_helpers.tpl
pmahindrakar-oss: This is not used , its already disabled. this is carried over from cluster-sync deployment which i haven't imported here as we dont run in self hosted as we dont run cloudflare tunnel here

charts/controlplane/values.yaml
pmahindrakar-oss: This is already in db block, can you elaborate more what you mean. 
I would move the value of the db into separate file for testing but kept for iteration if needed by others

charts/controlplane/values.yaml
pmahindrakar-oss: Yup, utlimately i hoping we remove these. Would get removed in next iterations

pmahindrakar-oss: Made these empty

charts/controlplane/values.yaml
pmahindrakar-oss: Moved it out and added empty

charts/controlplane/values.yaml
pmahindrakar-oss: called it artifacts-bucket for now

charts/controlplane/values.yaml
pmahindrakar-oss: There is global section which has most of the consolidated configs and only the overrides are present in each.
I have added a recursive helper for the prupose of merging these. At this point i find the configs are at the level where they are minimal atm and importing from generic, eg : db is not something all services need and i dont want to add those to global config

ctxswitch: I'm ok with that.  I think as we use it more we'll start to notice where some of the optimizations/consolidations could happen from an enduser perspective.  I'm more worried about creating a complex values file for the users that has too much redundancy - it makes the experience a little more cumbersome.

ctxswitch (approved):

charts/controlplane/values.yaml
pmahindrakar-oss: Since we have a way of consolidating things in global config it becomes easier to manage. Let me know any specific ones you found redundant.

pmahindrakar-oss: In the context of the database, I imagine we’ll either need to consolidate services that rely on the DB to reduce redundancy, or take the opposite approach and include the DB configuration in all services, regardless of whether they actually use it.

pmahindrakar-oss: Also the current consolidation in the global section doesn't exist in BYOC and we can iterate to get more things as we now have the process in place for it. Also we move things out of the global section which are completely static per deployment to reduce the configuration

PR #115, opened 2025-05-19
Upgrade knative-operator to 2025.5.0

mhotan:


charts/dataplane/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: dataplane
 description: Deploys the Union dataplane components to onboard a kubernetes cluster to the Union Cloud.
 type: application
 icon: https://i.ibb.co/JxfDQsL/Union-Symbol-yellow-2.png
-version: 2025.5.1
+version: 2025.5.2
```
pmahindrakar-oss, line 6: Assuming this is already working for app serving in BYOC ?

ctxswitch (approved):

charts/dataplane/Chart.yaml
mhotan: I am unclear of the question here. This patch should not change app-serving. This change is intended to fix existing fresh helm-chart as we moved knative crd into the dataplane but did not install by default until the most recent version of knative.

PR #116, opened 2025-05-20
Fix image builder helm-charts

mhotan:
* Bump dataplane version from `2025.5.2` -> `2025.5.3`
* Fix image builder buildkit deployment helm template to remove breaking
  condition when not using rootless
* Default to `non-rootless` buildkit to support a wider varient of out
  of the box kernels. For example EKS AutoMode using bottlerocket which
  limits user namespaces and breaks buildkit deployment
* add ephemeral-storage


mhotan:
fixes D-1350

ctxswitch (approved):

PR #129, opened 2025-06-11
Add ability to turn off dataplane billing

troychiu:


tests/generated/dataplane.aws.with-ingress.yaml
```diff
@@ -4206,11 +4164,6 @@ metadata:
     heritage: "Helm"
 spec:
   
-  sampleLimit: 0
-  targetLimit: 0
-  labelLimit: 0
-  labelNameLengthLimit: 0
-  labelValueLengthLimit: 0
```
pmahindrakar-oss, line 4213: Faced the same issue and helm tests would fail in CI.
Use the same version used by CI locally 
Recently upgrade to this due to the above issue
``` helm version
version.BuildInfo{Version:"v3.18.0
```

pmahindrakar-oss (approved):

PR #130, opened 2025-06-11
Usage ingress depending on billing

troychiu:


charts/controlplane/templates/common/_usage-ingress.yaml
```diff
@@ -0,0 +1,203 @@
...
+      name: usage
+      port:
+        number: 80
+- path: /cloudidl.usage.UsageService/ListQuotas/*
+  pathType: ImplementationSpecific
+  backend:
+    service:
+      name: usage
+      port:
+        number: 80
```
pmahindrakar-oss, line 59: This might be hard to maintain if we need to define every API as rule in ingress.
Have you considered excluding Billing specific paths which inverses what you have done

troychiu: Yeah that makes sense. Updated!

pmahindrakar-oss (approved):
Can we also add test with billing enabled , we can add that separate values file similar to this https://github.com/unionai/helm-charts/blob/cb946396fe52387b495cb47a16a7ae1692ae260b/tests/values/dataplane.aws.with-ingress.yaml

tests/run.sh
```diff
@@ -25,6 +25,7 @@ function generate {
     helm repo add opencost https://opencost.github.io/opencost-helm-chart
     helm repo add nvidia https://nvidia.github.io/dcgm-exporter/helm-charts
     helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
+    helm repo add flyte https://helm.flyte.org
```
troychiu, line 28: I have to add this, otherwise the test would fail

pmahindrakar-oss: Interesting . We already have core chart dependency. Will take a look at this later. You can merge this in

PR #131, opened 2025-06-11
Add scheduler and datacatalog services

pmahindrakar-oss:
Tested this on selfhosted and verified datacatlog and flytescheduler come up.


charts/controlplane/values.yaml
```diff
@@ -464,7 +461,7 @@ flyte:
       union:
         internalConnectionConfig:
           enabled: true
-          urlPattern: '{{ printf "_SERVICE_.%s.svc.cluster.local:80" .Values.namespace }}'
+          urlPattern: '{{ printf "_SERVICE_.%s.svc.cluster.local:80" .Release.Namespace }}'
```
mhotan, line 464: 👍

PR #143, opened 2025-06-18
Increase cpu requests for usage service

pmahindrakar-oss:


pmahindrakar-oss:
Yes i have seen the crash frequency reduced considerably after this , but again we are running on very small nodes and hence possibly the throttling. Making the current config sync and we can dig deeper on utlization

PR #152, opened 2025-08-30
K8s Plugin config and Union controller.

mhotan:
- Update Union operator config with Flyte propeller Kubernetes Plugin
  config so we can forward default labels, taints, and tolerations.
- Update helm-chart to conditionally deploy controller.

Depends on Union-operator release

- `jan/wip-selfhosted` - :warning: No PR associated with branch 
  - \#323
    - **K8s Plugin config and Union controller.** :point\_left:

### DIFF
--- a/charts/dataplane/templates/operator/serviceaccount.yaml
+++ b/charts/dataplane/templates/operator/serviceaccount.yaml
@@ -161,6 +161,14 @@ rules:
       - update
       - patch
       - delete
+  # Allow the controller check for existing CRDs before potentially creating controllers for them.
+  - apiGroups:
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yaml
```diff
@@ -23,9 +23,9 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Helm
-        uses: azure/setup-helm@v4.3.0
+        uses: azure/setup-helm@v4.3.1
         with:
-          version: v3.18.0
+          version: v3.18.6
```
mhotan, line 28: Had to update this to get helm validate to pass.

PR #210, opened 2026-01-26
:rotating_light: Breaking Change - Dataplane - Sane defaults

EngHabu:
\[:rotating\_light: Breaking Change]

Major overhaul of the dataplane chart to establish sane, low-privilege defaults that work out of the box for new deployments while preserving backward compatibility via values-legacy.yaml.

## Default mode changes

- Low-privilege mode (low\_privilege: true) is now the default — namespace-scoped RBAC, no cluster-wide permissions except where strictly required (fluentbit DaemonSet, knative-operator)
- V2 executor is the default — flytepropeller.enabled: false, executor.idl2Executor: true, flyteconnector.enabled: true
- Cluster resource sync disabled by default (clusterresourcesync.enabled: false)
- Common service account (union-system) shared across all components by default
- values-legacy.yaml created to restore all previous defaults with a single overlay file

## Prometheus consolidation

- Replaced the static prometheus deployment (templates/prometheus/) and the prometheus-simple subchart with a single community prometheus chart aliased as prometheus
- Removed standalone kube-state-metrics dependency (now a subchart of prometheus)
- Namespace-scoped RBAC for prometheus and kube-state-metrics in low-privilege mode via prometheus-rbac.yaml
- Scrape configs for flytepropeller, serving-envoy, and dcgm-exporter are now unconditional (no-op when targets don't exist)
- cAdvisor scraping moved to values-legacy.yaml (requires ClusterRole)

## Knative operator improvements

- Split CRDs into knative-operator-crds subchart — solves chicken-and-egg CRD validation error on fresh installs. CRDs are in the crds/ directory so Helm installs them before templates
- CRDs chart conditioned on knative-operator-crds.enabled
- Removed single\_namespace support — the operator requires cluster-scoped RBAC and can't be namespace-restricted
- Fixed \_example key in operator ConfigMaps that caused webhook validation failures on helm upgrade
- Added tpl support in knative-operator.namespace helper for namespaceOverride
- Fixed CRD conversion webhook namespace to use the helper instead of .Release.Namespace

## Image builder auto-configuration

- imageBuilder.defaultRepository auto-generates from cloud provider:
  - AWS: `<account_id>.dkr.ecr.<region>.amazonaws.com/<registryName>`
  - GCP: `<region>-docker.pkg.dev/<projectId>/<registryName>`
  - Azure: `<registryName>.azurecr.io`
- imageBuilder.authenticationType auto-detects (aws, google, azure, noop)
- New imageBuilder.registryName value (default: union-dataplane)
- Added global.AWS\_ACCOUNT\_ID for ECR URL generation

## Other changes

- Fixed fluentbit.serviceAccount.name to use common SA (union-system) by default
- Fixed union-serviceaccount.yaml missing tpl calls (pre-existing bug)
- Webhook templates moved from nodeexecutor/ to webhook/ directory with Helm-managed certificates
- Added GCP test case (dataplane.gcp.yaml)
- Removed values-low-privilege.yaml and values.v2.yaml (superseded by new defaults)

## Webhook certificate management

- Helm-managed MutatingWebhookConfiguration — the webhook configuration is now created by Helm (with flytepropellerwebhook.managedConfig: true) instead of self-registered by the webhook binary at runtime. This removes the need for the webhook to have cluster-scoped RBAC for mutatingwebhookconfigurations
- Configurable certificate providers (flytepropellerwebhook.certificate.provider):
  - helm (default) — generates self-signed certs using Helm's crypto functions, preserved on upgrade if the secret exists
  - certManager — uses cert-manager to provision and manage certificates
  - external — user-provided certificates via caCert, tlsCrt, tlsKey values
  - legacy — original behavior where the webhook binary generates its own certs via init container
- Static test certificates (values-test-certs.yaml) for deterministic test snapshot generation
- Webhook templates restructured from nodeexecutor/ to webhook/ directory

## :rotating\_light: Preserving current behavior

In order to preserve current behavior, `dataplane/values-legacy.yaml` is added to revert all defaults to current values. This can be used in conjunction with any other values file like this:

```
helm upgrade ..... -f values-legacy.yaml -f myvalues.yaml
```

## Test plan

- Verify helm template renders correctly for default values
- Verify helm template with low\_privilege: true produces namespace-scoped resources and no cluster-wide permissions
- Verify build-image-config configmap is created only in single-namespace mode
- Verify depot-token imagePullSecret appears in task pod template when Depot is enabled
- Verify webhook certificates render correctly for all providers (helm, certManager, external, legacy)
- Verify generated test fixtures match expected output
- Test upgrade path from existing deployments (webhook secret reuse, resource renames)

* `main` 
  - **:rotating\_light: Breaking Change - Dataplane - Sane defaults** :point\_left:
    - \#259


charts/dataplane/templates/propeller/service-webhook.yaml
davidmirror-ops:
```suggestion
  name: union-pod-webhook
```

davidmirror-ops (approved):
Not that testing is done but at least with the `singleNamespace` and `low_privilege` modes, the base components look and behave healthy

PR #333, opened 2026-04-10
Mitigate issues on upgrade to new defaults

davidmirror-ops:
Covers the following issues:

- Switch from union to community prometheus and mixed retry scenarios
- knative resources (CRDs) not being installed and causing failure to upgrade to `2026.4.5` 
- knative resources not being deleted on helm uninstall
- Bumps to `2026.4.6`

Introduces pre-upgrade and post-delete hooks to automate cleanup tasks

### DIFF
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -1843,7 +1843,7 @@ knative-operator-crds:
 knative-operator:
   enabled: true
   # Install into the release namespace rather than the default "knative-operator" namespace.
-  # Requires knative-operator >= 2025.6.4 which evaluates this via tpl.
+  # Requires knative-operator >= 2026.4.5 which evaluates this via tpl.
   namespaceOverride: '{{ .Release.Namespace }}'
   # Skip namespace creation — the parent chart creates the release namespace via --create-namespace
   skipNamespaceCreation: true
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -1475,6 +1475,14 @@ prometheus:
     emptyDir:
       sizeLimit: "10Gi"
 
+    # Explicitly use RollingUpdate and clear any rollingUpdate block to avoid
+    # the prometheus chart 25.x bug where rollingUpdate is set alongside Recreate.
+    strategy:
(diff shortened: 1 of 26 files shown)


EngHabu (approved):

charts/dataplane/values.yaml
```diff
@@ -1834,9 +1842,10 @@ knative-operator-crds:
 # Note: The operator requires cluster-scoped RBAC (CRDs, ClusterRoles, namespaces, deployments).
 knative-operator:
   enabled: true
-  namespaceOverride: '{{ .Release.Namespace }}'
```
EngHabu, line 1837: Why remove this? I intentionally wanted to force the namespace.. no?

charts/dataplane/values.yaml
```diff
@@ -1834,9 +1842,10 @@ knative-operator-crds:
...
   # Skip namespace creation — the parent chart creates the release namespace via --create-namespace
   skipNamespaceCreation: true
+  crds:
```
EngHabu: This is now split into a separate helm chart `knative-operator-crds`

davidmirror-ops: But customers would need to install that one first no? I want to prevent that

charts/dataplane/values.yaml
davidmirror-ops: What I found is that with this in place, the namespace was passed literally as `{{ .Release.Namespace }}` so some templating issue. Running as part of a `helm upgrade ... -n union...` it would just use that ns. Maybe not the most elegant fix I agree

charts/dataplane/values.yaml
EngHabu: It's added as a dependency for the dataplane chart, so they shouldn't need to

davidmirror-ops: without this it fails because the operator cannot find the `certificates.networking.internal.knative.dev` CRD. 

I saw this the 2 times I tried but can take a closer look

davidmirror-ops: Fixed with #335

EngHabu (approved):

PR #345, opened 2026-04-17
Revert duplicated CRDs

davidmirror-ops:
Fixes CRD ownership conflicts that broke helm install and helm upgrade when app serving (knative-operator) is
enabled.

Root cause: The knative-operator-crds subchart installed knativeeventings and knativeservings CRDs via its crds/
directory (no Helm ownership labels), while templates/serving/knative-operator-crds.yaml attempted to manage the
same two CRDs as tracked release resources (which requires ownership labels). This caused two distinct failures:

- Install: CustomResourceDefinition "knativeeventings.operator.knative.dev" exists and cannot be imported — Helm
  refused to take ownership of CRDs it didn't originally track
- Upgrade: no matches for kind "KnativeServing" — Helm's crds/ directory is skipped on upgrade, so the CRDs were
  absent when the post-upgrade hook ran

Fix:

- Remove knative-operator-crds as a subchart dependency (from Chart.yaml and values.yaml) — it was the source of
  both failures
- Retain templates/serving/knative-operator-crds.yaml with only the two operator CRDs (knativeeventings,
  knativeservings) as proper Helm-tracked resources with helm.sh/resource-policy: keep; this ensures they are
  applied on every install and upgrade, and are never deleted on uninstall
- The remaining 12 Knative Serving CRDs (routes, revisions, services, etc.) are installed by the knative-operator
  itself when it reconciles the KnativeServing CR — no pre-installation needed

Test plan

- helm install with serving.enabled=true succeeds on a fresh cluster
- helm upgrade from a prior version with serving.enabled=true succeeds
- helm uninstall leaves the two operator CRDs in place (resource-policy: keep)
- make test passes

* `main` 
  - **Revert duplicated CRDs** :point\_left:
    - \#347


mhotan (approved):

EngHabu (approved):
Tested:
```
Verification: 7/7 passed
  Live & Persistent Logs                   PASSED
  Inputs/Outputs                           PASSED
  Code Bundle Download & CORS              PASSED
  Remote Image Builder                     PASSED
  Image Cache (No Rebuild)                 PASSED
  Reusable Containers                      PASSED
  App Deployment                           PASSED
```

PR #357, opened 2026-04-22
Add flyteidl2.cluster.ClusterService ingress route for SDK v2

mhotan:
## Summary
- Adds `/flyteidl2.cluster.ClusterService` and `/flyteidl2.cluster.ClusterService/*` routes to the protected gRPC ingress, targeting the `cluster` service on the `connect` port
- Without this, the flyte-sdk v2 `SelectCluster` call falls through to the console catch-all and gets `text/html` instead of `application/proto`
- Route already exists in BYOC (dogfood `helm-protected-grpc` ingress) but was missing from the selfhosted controlplane chart

## Test plan
- [ ] Verify `flyte run` works on identity-testing after deploy

### DIFF
--- a/charts/controlplane/templates/common/_ingress-protected.yaml
+++ b/charts/controlplane/templates/common/_ingress-protected.yaml
@@ -393,6 +393,20 @@
       name: cluster
       port:
         name: connect
+- path: /flyteidl2.cluster.ClusterService/*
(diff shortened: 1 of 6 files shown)


katrogan (approved):

katrogan:
oh wait @mhotan don't we need to bump the base cloud image?

PR #360, opened 2026-04-22
Fix/cleanup stale/immutable resources

davidmirror-ops:
This change aims to address the following issues during upgrades from pre `2026.4.7` versions

1. Immutable Deployment Selector Change                                                      

  The chart renamed the `flytepropeller-webhook` deployment's selector label from flyte-pod-webhook to
   union-pod-webhook. Kubernetes does not allow changing spec.selector on an existing Deployment — the deployment must be deleted and recreated. ArgoCD's ServerSideApply sync strategy cannot handle this; a --replace sync is required for this specific resource.

  Affected resource: Deployment/union-pod-webhook
  Error: spec.selector: Invalid value: field is immutable

  2. Immutable Secret

  The `flyte-pod-webhook` Secret is marked as `immutable: true.` The new chart version regenerates
  this secret with new TLS certs, but Kubernetes rejects updates to immutable secrets. The
  secret must be deleted before syncing.

  Affected resource: Secret/union-pod-webhook
  Error: data: Forbidden: field is immutable when immutable is set

  3. MutatingWebhookConfiguration Name Change

  The chart renamed the MutatingWebhookConfiguration from union-pod-webhook to
  union-pod-webhook-mistral. This leaves the old webhook config orphaned with a stale CA
  bundle, causing TLS verification failures when pods are created in flyte-default.

  Affected resource: MutatingWebhookConfiguration/union-pod-webhook
  Error: tls: failed to verify certificate: x509: certificate signed by unknown authority
  Fix: Delete the old union-pod-webhook MutatingWebhookConfiguration manually.

4. Immutable selector on union-operator-prometheus
```
Deployment: Deployment.apps "union-operator-prometheus" is     
  invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"prometheus",     
  "app.kubernetes.io/instance":"ml-orchestrator-dp", "app.kubernetes.io/name":"ml-orchestrator-dp-dataplane"},                          
  MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

### What this change does

It introduces a pre-upgrade hook that will delete these 4 resources before upgrading the release:

1. flytepropeller-webhook deployment
2. flyte-pod-webhook secret
3. flyte-pod-webhook MutatingWebhookConfiguration
4. `union-operator-prometheus` deployment

Cleans temporary SA and ClusterRoles on hook success or fail.

It only acts on upgrades, not installs.

Once our customer base is migrated we can retire this component.

### DIFF
--- a/charts/dataplane/templates/webhook/pre-upgrade-cleanup.yaml
+++ b/charts/dataplane/templates/webhook/pre-upgrade-cleanup.yaml
@@ -16,7 +16,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     helm.sh/hook: pre-upgrade
-    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
     helm.sh/hook-weight: "-5"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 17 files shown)


EngHabu (approved):

davidmirror-ops:
Merging as I was able to upgrade and test executions (reusable and app serving) successfully.

PR #362, opened 2026-04-27
Skip default registries from KNative digest check

EngHabu:


tests/generated/dataplane.oci.yaml
```diff
@@ -10011,7 +10011,7 @@ spec:
       queue-sidecar-memory-limit: "800Mi"
       queue-sidecar-ephemeral-storage-request: "512Mi"
       queue-sidecar-ephemeral-storage-limit: "1024Mi"
-      registries-skipping-tag-resolving: managed.cr.union.ai
+      registries-skipping-tag-resolving: managed.cr.union.ai,ghcr.io,union-dataplane
```
jeevb: Do the default registries maybe need the docker.io/library/ prefix or leave it up to users?

jeevb (approved):

PR #364, opened 2026-04-28
Set db.debug to false in controlplane values

chloe-kx:
## Summary
- Explicitly set `db.debug: false` in all controlplane `db:` config sections (artifacts, cluster, organizations, executions, run-scheduler)
- Ensures GORM SQL debug logging is disabled by default across all services
- Related to unionai/cloud#15595 which decouples GORM logging from app log level

## Test plan
- [ ] CI snapshot tests pass
- [ ] Verify rendered ConfigMaps contain `debug: false` under `db:` sections


charts/controlplane/values.yaml
```diff
@@ -410,6 +410,7 @@ services:
         metrics:
           scope: "artifacts:"
       db:
+        debug: false
```
mhotan: Apologies this is not obvious. But there should be some baked in helm logic that allows us to configure this in one location and appropriately propagate to each service.

mhotan (approved):

PR #376, opened 2026-05-07
Add organizations service bootstrap for selfhosted deployments

jmonty42:
## Summary
- Adds bootstrap init container to organizations service that runs `bootstrap-tenant organizations` after DB migration
- Adds `bootStrapConfig` to organizations configMap using `global.UNION_ORG`

## Problem
The organizations service requires a seeded `orgs` table for the identity service prefetcher to list users. Without this bootstrap, `OrgService/List` returns 0 orgs and the identity service reports "Prefetched 0 users/apps" — users don't appear in the UI.

Unlike the authorizer (which has automatic `RunBootstrap` in `GrpcRegistrationHook`), the organizations service requires the CLI command `cloudorganizations bootstrap-tenant organizations` to be explicitly invoked.

## Test plan
- [x] `make generate-expected && make test` passes
- [x] Deploy to monty-selfhosted and verify orgs table is populated
- [ ] Verify users appear in the UI after deployment

### DIFF
--- a/charts/controlplane/values.yaml
+++ b/charts/controlplane/values.yaml
@@ -763,6 +763,13 @@ services:
           - migrate
           - --config
           - "/etc/config/*.yaml"
+      - name: bootstrap
+        args:
+          - cloudorganizations
+          - bootstrap-tenant
+          - organizations
+          - --config
+          - "/etc/config/*.yaml"
     args:
       - cloudorganizations
       - serve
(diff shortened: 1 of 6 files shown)


charts/controlplane/values.yaml
```diff
@@ -763,6 +763,13 @@ services:
           - migrate
           - --config
           - "/etc/config/*.yaml"
+      - name: bootstrap
+        args:
+          - cloudorganizations
+          - bootstrap-tenant
```
mhotan, line 769: I did not know this was a thing.

mhotan (approved):

PR #386, opened 2026-05-12
Opt-in Zero Trust

jeevb:
Add ability for users to opt-in to zero-trust in their clusters, simply by including the `values.zero-trust.yaml` overlay.

- `main` 
  - **Opt-in Zero Trust** :point\_left:
    - \#422


charts/dataplane/templates/operator/configmap.yaml
```diff
@@ -29,8 +29,23 @@ data:
       enableTunnelService: {{ .Values.config.operator.enableTunnelService }}
       #  enableDepot: {{ include "operator.enableDepot" . | default "false" }}
       tunnel:
-        enableDirectToAppIngress: {{ .Values.serving.enabled }}
+        enableDirectToAppIngress: {{ and .Values.serving.enabled .Values.tunnelService.apps.enableGlobalIngress }}
```
EngHabu, line 32: `or`?

EngHabu (approved):
Haven't looked at the bulk of the changes. Only where operator config/values changed... Looks good to me minus one comment

EngHabu:
Would like @mhotan and @davidmirror-ops to take a look too

charts/dataplane/templates/operator/configmap.yaml
jeevb: This is the flag that enables the `apps.*` route in the tunnel config. I think `and` is right because we don't want to enable it if serving is disabled, or apps global ingress is disabled.

EngHabu: Don't we always want apps.* route? if either serving is on (legacy no zero trust) or zero trust

jeevb: Ah, until we have multi-cluster support for apps, I figured this would be a useful toggle to have. So in the secondary cluster, you just disable global ingress, but keep serving on, so that we could have the escape hatch of still using the cluster-specific URLs. `.Values.tunnelService.apps.enableGlobalIngress` defaults to `true`, so should be a no-op for existing chart users.

charts/dataplane/templates/gateway/activator.yaml
```diff
@@ -0,0 +1,242 @@
+{{- if .Values.gateway.enabled }}
```
mhotan: Can I ask what version of Knative Serving / Operator this is? I am little curious of the maintainance trade off here. Vendoring Knative Serving resources will make it difficult and confusing to update for anyone outside a small set of people.

charts/dataplane/templates/dataproxy/deployment.yaml
```diff
@@ -0,0 +1,84 @@
+{{- if .Values.dataproxy.enabled }}
```
mhotan: Any reason we don't enable by default?

charts/dataplane/values.yaml
```diff
@@ -2301,3 +2315,290 @@ monitoring:
         requests:
           cpu: "500m"
           memory: "1Gi"
+
+# -- Vendored Knative Serving + Kourier + Envoy gateway. BYOC zero-trust mode only.
```
mhotan:
```suggestion
# -- Vendored Knative Serving + Kourier + Envoy gateway. Zero-trust mode only.
```

charts/dataplane/templates/dataproxy/deployment.yaml
jeevb: Zero-trust is opt-in today to not break existing users of the chart. Existing users will see a small change to the operator config, which is mostly a no-op anyway. The goal is to keep this until we migrate all existing customers over, and then we can simplify.

charts/dataplane/templates/gateway/activator.yaml
jeevb: It is the exact same version we're using today, installed by Knative-Operator 1.16.0. Figured parity first, then update as necessary.

charts/dataplane/templates/operator/configmap.yaml
```diff
@@ -29,8 +29,23 @@ data:
...
         deploymentToRestart: union-operator-proxy
+        {{- if .Values.tunnelService.apps.originURL }}
+        appIngressOriginURL: {{ .Values.tunnelService.apps.originURL | quote }}
+        {{- else if .Values.gateway.enabled }}
+        appIngressOriginURL: "http://{{ include "gateway.fullname" . }}:{{ .Values.gateway.service.httpAppsPort }}"
```
mhotan, line 37: Cool so vendored gateway service and act as an Ingress for Apps, so we can replace kourier-internal which I believe comes from Knative Operator.

mhotan (approved):
I'd like to merge https://github.com/unionai/cloud/pull/16015 before this and create a PR ontop of this to move vendored CRDs into a non chart directory. It would be great to to do prior to merging this PR because

PR #387, opened 2026-05-12
dataplane: disable 2nd mutating webhook config when managedConfig is set

EngHabu:
## Summary

The dataplane chart sets `disableCreateMutatingWebhookConfig: true` (in both the `union-pod-webhook` helper and the propeller configmap) only when `low_privilege` is set. But when `flytepropellerwebhook.enabled` + `flytepropellerwebhook.managedConfig` are set, the webhook deployment manages its own `MutatingWebhookConfiguration` — so propeller should not also try to create one, or we end up with a second/conflicting config.

This gates the `disableCreateMutatingWebhookConfig: true` setting on either condition:

```
{{- if or .Values.low_privilege (and .Values.flytepropellerwebhook.enabled .Values.flytepropellerwebhook.managedConfig) }}
```

Applied in both places that previously checked `low_privilege` alone:
- `charts/dataplane/templates/_helpers.tpl` (union-pod-webhook helper)
- `charts/dataplane/templates/propeller/configmap.yaml`

## Test plan

- [x] `make generate-expected` — `tests/generated/dataplane.clusterresourcesync.yaml` picks up `disableCreateMutatingWebhookConfig: true` in the propeller webhook config + corresponding `configChecksum` updates
- [x] `make test` passes
- [x] Validate on a selfhosted dataplane with `flytepropellerwebhook.enabled=true` + `managedConfig=true` that only one `MutatingWebhookConfiguration` is reconciled

### DIFF
--- a/charts/dataplane/templates/_helpers.tpl
+++ b/charts/dataplane/templates/_helpers.tpl
@@ -1548,7 +1548,7 @@ union-pod-webhook
 {{- $_ := set $webhook "serviceName" (include "flytepropellerwebhook.serviceName" .) }}
 {{- $_ := set $webhook "secretName" (include "flytepropellerwebhook.secretName" .) }}
 {{- $_ := set $webhook "localCert" true }}
-{{- if .Values.low_privilege }}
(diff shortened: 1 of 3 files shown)


davidmirror-ops (approved):
Thank you!

PR #388, opened 2026-05-12
Revert imageBuilder authenticationType auto-detect

mhotan:
## Summary

Reverts the imageBuilder authentication-type auto-detection introduced in PR #210 ([commit `cb48bbcd`](https://github.com/unionai/helm-charts/commit/cb48bbcdbb544c4c2598bb480dbcc5b55add5fb2)). The configmaps now render the explicit `.Values.imageBuilder.authenticationType` directly, so an operator's configured value (including `"noop"`) is honored verbatim.

## Why

A selfhosted operator reported that their explicit `imageBuilder.authenticationType: "noop"` on AWS was being silently rewritten to `"aws"` by the helper, breaking their ECR auth setup.

- Helper introduced in: PR #210 "🚨 Breaking Change - Dataplane - Sane defaults" (commit [`cb48bbcd`](https://github.com/unionai/helm-charts/commit/cb48bbcdbb544c4c2598bb480dbcc5b55add5fb2))

The helper treated `"noop"` as a sentinel for "auto-detect", which prevented operators from intentionally selecting that mode on a recognized cloud. Explicit is better — operators who want the provider credential chain can set the value themselves.

## Changes

- Remove the `imagebuilder.authenticationType` helper from `charts/dataplane/templates/_helpers.tpl`
- Render `.Values.imageBuilder.authenticationType` directly in:
  - `charts/dataplane/templates/imagebuilder/build-image-configmap.yaml`
  - `charts/dataplane/templates/operator/configmap.yaml`
- Add a commented-out hint in each cloud overlay (`values.{aws,gcp,azure}*.yaml`) pointing operators at the explicit `authenticationType` value if they want the provider credential chain
- Regenerate snapshots — AWS/GCP/Azure tests now render `"noop"` (chart default) instead of the auto-derived cloud value

## Migration

Operators who relied on the auto-detection (i.e., didn't set `imageBuilder.authenticationType`) and need the cloud provider credential chain must now set the value explicitly in their values file:

```yaml
imageBuilder:
  authenticationType: aws    # or "google" / "azure" / "basic"
```

## Test plan

- [x] `make generate-expected` regenerated snapshots
- [ ] CI `make test` passes
- [ ] Operator confirms `"noop"` is now respected on their deployment

### DIFF
--- a/charts/dataplane/templates/_helpers.tpl
+++ b/charts/dataplane/templates/_helpers.tpl
@@ -1493,25 +1493,6 @@ Otherwise, build it from imagebuilder.defaultRegistry plus the provider-specific
 {{- end -}}
 
 {{/*
-Returns the image builder authentication type.
-If imageBuilder.authenticationType is explicitly set (non-empty, not "noop"), use it.
-Otherwise, auto-detect from the cloud provider.
-*/}}
-{{- define "imagebuilder.authenticationType" -}}
-{{- if and .Values.imageBuilder.authenticationType (ne .Values.imageBuilder.authenticationType "noop") -}}
-  {{- .Values.imageBuilder.authenticationType -}}
-{{- else if eq (tpl .Values.storage.provider .) "aws" -}}
-  {{- "aws" -}}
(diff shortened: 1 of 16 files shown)


davidmirror-ops (approved):
ah that makes sense. Thanks

PR #392, opened 2026-05-15
Rerun make generate-expected

jeevb:


mhotan (approved):
Feel free to update to latest Helm in github actions config: https://github.com/unionai/helm-charts/blob/main/.github/workflows/checks.yaml#L19-L32

PR #399, opened 2026-05-20
Point dataplane to upstream subchart again

davidmirror-ops:
Changes applied previously to mitigate upgrade issues made necessary to bump the knative-operator subchart version. To avoid things breaking while the release completed (which happens on merge) we temporarily point to the internal folder.


EngHabu (approved):

EngHabu:
Is this the right version to point to? should we bump that?

PR #411, opened 2026-06-01
Changes to enable billing collection in low priv mode

katrogan:


charts/dataplane/values.yaml
```diff
@@ -540,6 +540,10 @@ config:
       userRole: "{{ tpl (.Values.userRoleAnnotationValue | toString) $ }}"
       # -- For GCP only, the project id for object storage.
       gcpProjectId: "{{ tpl .Values.storage.gcp.projectId . }}"
+      # -- Cloud provider the cluster runs on (aws, gcp, azure, oci, or metal). Selects the GPU
+      # accelerator node label used for usage attribution and tags billable usage. Defaults to the
+      # top-level `provider`. When empty, the operator falls back to detecting it from node labels.
+      cloudProvider: '{{ tpl (.Values.provider | default "" | toString) . }}'
```
davidmirror-ops, line 547: How does it fallback? Does it try to infer the device from the injected nodeSelector labels?

davidmirror-ops (approved):

charts/dataplane/values.yaml
katrogan: yes exactly! so the fallback is only enabled when low priv isn't set. let me update the comment

PR #420, opened 2026-06-03
dataplane: set proxy smConfig.webhookHostTemplate default

EngHabu:
## Overview
Sets the default for the new dataproxy `SecretManagerConfig.WebhookHostTemplate` field (added in unionai/cloud#16293) to `union-pod-webhook-headless.%s.svc`.

The dataproxy secret-manager service uses the webhook headless service to invalidate the secret cache across all webhook replicas. This chart renders the headless webhook service as `union-pod-webhook-headless` (`{{ template "flyte-pod-webhook.name" . }}-headless`), so the template is set accordingly. `%s` is replaced at runtime with the operator pod namespace.

## Test Plan
- Value matches the headless `Service` name produced by `charts/dataplane/templates/webhook/service.yaml`.

## Related
- unionai/cloud#16293

### DIFF
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -537,6 +537,8 @@ config:
       enabled: "{{ .Values.proxy.secretManager.enabled }}"
       # -- The type of secret manager to use.
       type: "{{ .Values.proxy.secretManager.type }}"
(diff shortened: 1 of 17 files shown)


davidmirror-ops (approved):
Oh I see, thank you!

PR #425, opened 2026-06-04
Actions leasor

wild-endeavor:
## Data Plane changes

Two unused leaseworker config values removed.

## Control Plane changes

Values added:

- Dataplane cluster name.
- Two new configuration options for run service
- Leasor deployment with config map.
- Top level actions key and associated values for the actions service.

### Defaults

- `useActionsServiceForOrgs: []` and `rejectLegacySDKVersions: false` by default.
- `alpine/k8s:1.32.3` init image promoted to `actions.coordination.image.*` for air-gapped / customer-mirrored installs.

### Actions service

Add the actions service and the associated actions router.

- Actions is sharded by default so coordinator to help with shard handoff is included.
- Router image will need to be published - this is the new `envoy-router` image.

### Leasing service

- Router is not yet included, can be included in the future once the sharding vector is decided and implemented.  So currently leasor is just a generic service.
- Leasor needs a custom config map because it needs to create queue settings based on the dataplane cluster name.  Assumed to be only one right now - can revisit as part of multi-dataplane air-gapped work.

### Additional

- Ingress routes added.

* `main` 
  - **Actions leasor** :point\_left:
    - \#433


charts/controlplane/templates/actions/_helpers.tpl
```diff
@@ -0,0 +1,184 @@
+{{/*
+Actions service helpers. Actions is sharded + fronted by an Envoy router, so it
+cannot use the generic .Values.services renderer; these helpers and the
+templates/actions/* manifests reproduce that topology while reusing the chart's
+shared conventions (image fallback, config merge, labels).
+*/}}
+
+{{- define "actions.name" -}}
+actions
```
mhotan, line 9: Should be fine for now. Eventually we will move toward more common Helm conventions to prefix with {{ .ReleaseName }} (or something like that.

mhotan (approved):

PR #429, opened 2026-06-05
FAB-395: ci: add dataplane integration test on k3d with RustFS

EngHabu:
## Summary

- Adds a GitHub Actions workflow (`.github/workflows/dataplane-integration.yaml`) that spins up a k3d cluster (on the 4-vCPU `ubuntu-latest` runner), deploys RustFS as S3-compatible storage, installs the dataplane chart from the PR branch, registers it against the shared BYOK staging control plane (`byok.us-west-2.union.ai`), and runs a 6-scenario smoke suite end-to-end
- Uses `BYOK_URL` / `BYOK_CLIENT_ID` repo variables and `BYOK_CLIENT_SECRET` secret for auth — no hardcoded credentials, and these are never echoed to the log
- Creates a dedicated cluster pool + project per PR run (keyed by run id + attempt) so parallel PRs / re-runs never share a cluster, and sets up project→pool routing
- Tears down (deregisters cluster, deletes k3d) on completion regardless of outcome

## Smoke tests

After the cluster registers healthy and capabilities propagate, the workflow runs `hello` (baseline run, also feeds logs/io), then these six scenarios (imported from the shared smoke-test definitions). The job fails if any one fails:

| Test | What it verifies |
|---|---|
| `verify_logs` | Task stdout/stderr logs are captured and retrievable from the run |
| `verify_io` | Task inputs/outputs round-trip through the configured (RustFS) blob store |
| `verify_image_builder` | In-cluster image build (buildkit → k3d registry) succeeds and the task runs on the built image |
| `verify_image_cache` | A second build of the same image hits the cache instead of rebuilding |
| `verify_reusable` | A reusable (persistent) task environment accepts and serves multiple invocations |
| `verify_app` | A FastAPI app deploys via Knative serving, is assigned to this run's cluster (`status.assigned_cluster`), and answers its `/` and `/health` endpoints |

Scheduling notes for the 4-vCPU runner: the light tests run in parallel; `verify_reusable` and `verify_app` run sequentially (each needs a persistent pod that can't co-schedule with the other). Each test is bounded by a timeout so a stuck run fails rather than hanging the job.

## New files

| File | Purpose |
|---|---|
| `.github/workflows/dataplane-integration.yaml` | Main workflow |
| `.github/ci-scripts/ci_dataplane.py` | Python helper: provision, health poll, routing, smoke-suite driver, teardown |
| `.github/ci-values/dataplane.yaml` | CI values overlay (RustFS storage, trimmed Knative serving, disable monitoring/fluentbit/etc.) |
| `.github/ci-manifests/rustfs.yaml` | In-cluster RustFS deployment + service |
| `ci_smoke_task.py` | Task definitions for hello + the light smoke scenarios |
| `ci_app_task.py` | FastAPI app task definition for `verify_app` |

## Test plan

- [x] CI run passes with cluster reaching `STATE_ENABLED`
- [x] Cluster pool + project created for this run, with project→pool routing
- [x] All six smoke tests pass (6/6 green)
- [x] Teardown removes cluster registration

fixes https://linear.app/unionai/issue/FAB-395/test-dataplane-helm-chart-in-pr


asdasfasf121:
hey, sorry to hear about having to deal with those ridiculously low DeepSeek rate limits - it's frustrating af trying to make progress when you're constantly hitting them.

you know, i actually built an open-source proxy/pool called lucky that kind of aggregates DeepSeek keys and stuff, which basically guarantees 100% uptime since it dynamically adjusts the load.

i've got some free trial keys lying around if you wanna test it out and see if that solves your problem. it's not gonna magically fix everything, but it might at least make life a bit easier. check out the latest version on https://lucky.wangzhonghao.me and let me know if you have any questions.

byok-values.yaml
```diff
@@ -0,0 +1,24 @@
...
+  accessKey: <S3_ACCESS_KEY>
+  secretKey: <S3_SECRET_KEY>
+  bucketName: <S3_BUCKET_NAME>
+  fastRegistrationBucketName: <S3_BUCKET_NAME>
+  region: <CLOUD_REGION>
+secrets:
+  admin:
+    create: true
+    clientId: byok-ci-helm-test-preview-operator
+    clientSecret: AcUpSHiybZZ0BxBVruJPFswnu9w2d2HkCy3wykjy9WRoCexQ8HRsamPpQtJT7fPa
```
davidmirror-ops: should we check in this?

.github/ci-manifests/rustfs.yaml
```diff
@@ -0,0 +1,67 @@
...
+    matchLabels:
+      app: rustfs
+  template:
+    metadata:
+      labels:
+        app: rustfs
+    spec:
+      containers:
+        - name: rustfs
+          image: rustfs/rustfs:latest
```
davidmirror-ops: should we pin this to a particular digest/version?

.github/ci-scripts/ci_dataplane.py
```diff
@@ -0,0 +1,930 @@
...
+            "[ci] smoke-suite FAILED: "
+            + ", ".join(n for n, _ in failed)
+        )
+
+
+# ── teardown ────────────────────────────────────────────────────────────────
+
+def cmd_teardown(args: argparse.Namespace) -> None:
+    cluster_name = _env("CLUSTER_NAME")
+    print(f"[ci] teardown: deregistering cluster {cluster_name}", flush=True)
```
davidmirror-ops, line 896: @EngHabu seems like this only deletes the cluster but not the clusterpool, assignment and attributes that are uniquely keyed by run id

EngHabu: Unfortunately no APIs AFAIK for removing clusterpools or projects (there is archive)

davidmirror-ops: ah I see, hoping this doesn't bloat the CP at some point

byok-values.yaml
EngHabu: Good catch — no, we should not. That file was an unreferenced leftover (nothing in the repo consumes it) and it had a plaintext operator `clientSecret` in it. Removed in cb5fd63. It only ever existed on this branch, never on `main`. The exposed secret is being rotated separately on our side.

.github/ci-manifests/rustfs.yaml
EngHabu: Done — pinned to `rustfs/rustfs:1.0.0-beta.8@sha256:fa19210a…` (the current `:latest`) in cb5fd63, so the storage backend cant shift under a PR mid-run. Well bump it deliberately going forward.

.github/ci-scripts/ci_dataplane.py
EngHabu: You were right that it accumulated — teardown only ran `delete cluster`, leaking the per-run pool, assignment, 3 routing attributes and project every run. Fixed in cb5fd63: teardown now also deletes the clusterpool-attributes (dev/staging/prod), the assignment and the pool, and archives the project (Flyte has no project delete). All best-effort so cleanup never fails the `always()` step. Verified each `uctl delete` subcommand + signature against uctl v0.1.20 before wiring it up.

davidmirror-ops (approved):

EngHabu:
Heads up on the chart changes in this PR — I've split the non-CI bits out into #441 (vendored knative operator CRDs, buildkit `extraConfig` knob, operator configmap filter, and the regenerated dataplane snapshots). That's ~6.6k of the lines here.

Plan: merge #441 first; once it lands on main I'll sync this branch and those files drop out of this diff automatically, leaving #429 as a CI-only change (`.github/` + the CI task scripts). Keeping them here for now so this PR's integration test still passes until #441 merges.

PR #428, opened 2026-06-05
dataplane: make CP connection TLS-mode flags consistent + drop redundant overlay blocks

mhotan:
## Summary

Make the TLS-mode flags (`insecure`, `insecureSkipVerify`, `_U_INSECURE`, `_U_INSECURE_SKIP_VERIFY`) consistent across the four DP→CP connection consumers and flip the default to TLS-on. Drop redundant connection blocks from the cloud overlays since they were just restating values that already exist in base `values.yaml`.

## Why

Five places in the chart wire DP→CP gRPC connections:

| Consumer                                              | File            | TLS flag pair                            |
| ----------------------------------------------------- | --------------- | ---------------------------------------- |
| `config.admin.admin`                                  | base + overlays | `insecure`, `insecureSkipVerify`         |
| `config.catalog.catalog-cache`                        | base + overlays | `insecure`, `insecure-skip-verify`       |
| `config.union.connection`                             | base + overlays | `insecure`, `insecureSkipVerify`         |
| `clusterresourcesync.config.union.connection`         | base + overlays | `insecure`, `insecureSkipVerify`         |
| `config.k8s.plugins.k8s.default-env-vars` (task pods) | overlays only   | `_U_INSECURE`, `_U_INSECURE_SKIP_VERIFY` |

Base `values.yaml` had `insecure: false`. The AWS + GCP overlays restated each connection block with `insecure: true` (commented "intra-cluster HTTP"). The task-pod env vars set `_U_INSECURE: true`. The four control-plane defaults disagreed with each other and the task-pod env-var pair disagreed within itself (`_U_INSECURE: true, _U_INSECURE_SKIP_VERIFY: false` is nonsense — skipVerify only matters with TLS on).

That mismatch is the load-bearing bug. After cloud #16265 made the canonical DP→CP path `dns:///{CONTROLPLANE_HOST}:443` (TLS-terminating nginx on :443), `insecure: true` cleanly produces:

```
rpc error: code = Unavailable desc = connection error: desc =
"error reading server preface: http2: failed reading the frame payload:
http2: frame too large, note that the frame header looked like an HTTP/1.1 header"
```

That's `union-syncresources` (and any other consumer using the operator-side connection block) sending plain HTTP/2 to a TLS endpoint.

## Changes

[...]

PR #431, opened 2026-06-05
dataplane: unified top-level secrets block (eagerApiKey, eagerClientCreds, imageBuilder.{push,pull}, admin)

mhotan:
## Summary

Reshape the dataplane chart's secret provisioning into a single declarative `secrets:` block at the top level. Replaces three post-install side channels (`flyte create secret EAGER_API_KEY`, `flyte create secret image-builder-secret`, `kubectl patch depot-token`) with chart-managed Secrets that respect an `existingSecret.name` redirect for users wiring their own secret manager (Whisper / sealed-secrets / external-secrets).

Four commits, each defaults-off so generated test snapshots remain byte-identical with main until an environment opts in:

| Commit    | Description                                                                                                                                                                                                                                                                                                                                                          |
| --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `d399219` | Scaffold `templates/_secrets.tpl` consumer helpers (`secretRef`, `secretName`, `enabled`, `validate`) plus the unified `secrets:` block in `values.yaml`. No new templates, no behaviour change.                                                                                                                                                                     |
| `89eb7dd` | Materialize four templates (`eager-api-key`, `eager-client-creds`, `image-builder-push`, `image-pull`) when enabled. `task-podtemplate.yaml` consumes the new image-pull helper, falls through to the legacy `depot-token` reference when the helper is disabled. Each Secret carries a `union.ai/secret-type:` label so operator cleanup informers can identify it. |
| `ee35a99` | Normalize `secrets.admin` to the unified `existingSecret.name` + `enabled` shape while keeping the legacy `enable` / `create` fields as a deprecation shim. Default computed name preserved so existing deployments see no name change.                                                                                                                              |
| `7387df1` | Convert `operator-cluster-name` from Secret to ConfigMap (it is an identifier, not a credential). Same metadata.name preserved; `_helpers.tpl` consumer flipped from `secretKeyRef` to `configMapKeyRef`.                                                                                                                                                            |

## Why

Every selfmanaged environment has had to run post-install workarounds to populate these Secrets. Apple in particular wants declarative provisioning: either chart-managed Secrets from values, or references to pre-existing names. The new shape supports both via `existingSecret.name`. The `operator-cluster-name` change is hygiene — cluster names are configuration, not credentials, and shouldn't live in Secrets.

## Migration

See `charts/MIGRATION.md` for the per-secret migration table. TL;DR for existing deployments:

- All new fields default `enabled: false` → no behaviour change unless opted in
- `secrets.admin.enable` / `create` legacy fields still honoured (with deprecation warning) so existing values overlays keep working
- `operator-cluster-name` Secret → ConfigMap is the one resource-type flip in this PR — on helm upgrade / argocd sync, the existing Secret is removed and the ConfigMap created at the same name

## Test plan

- [x] `make test` — generated snapshots match expected
- [x] `helm template` rendering against three overlays (defaults-off, inline-value, existingSecret redirect)
- [x] mike-apple-aws (staging): ExternalSecret materializes K8s Secret, chart references `existingSecret.name`, PodTemplate carries the new imagePullSecrets entry
- [ ] End-to-end pull validation — task pod actually fetches a private-registry image via the chart-managed pull secret (label-mirror alignment pending verification)
- [ ] Push path validation — companion `build_image_task.py` + bootstrap-Job edits not yet in this PR (planned follow-up commit)

## Stacked on

- Base: `mike/dataplane-shared-connection-config` (the DP→CP TLS-default flip + connection wiring dedup). Will rebase onto main once that branch merges.

## Open

- Label on chart-managed Secrets is `union.ai/secret-type: image-builder-pull` etc.; existing operator mirror code may look for `union.ai/secret-type: image-pull-secret`. Need to align — either chart adjusts to legacy label or operator/webhook side adds parameterized label-set support (in the companion cloud-side PR).
- `build_image_task.py` push-secret volume mount + bootstrap Job env passthrough is the next commit to add here, enabling `Image.from_debian_base(registry="<private-registry>", name="<repo>")` to work without per-task `registry_secret`.

* `main` <!-- branch-stack -->
  - \#428
    - **dataplane: unified top-level secrets block (eagerApiKey, eagerClientCreds, imageBuilder.{push,pull}, admin)** :point\_left:

### DIFF
--- a/charts/dataplane/values.yaml
+++ b/charts/dataplane/values.yaml
@@ -1887,6 +1887,51 @@ secrets:
     # -- The client id used to authenticate to the control plane.  This will be provided by Union.
     clientId: dataplane-operator
 
+  # -- Eager-execution API key (DP → CP callback). Replaces post-install `flyte create secret EAGER_API_KEY`.
+  eagerApiKey:
+    # -- Reference an externally-provisioned Secret (Whisper, sealed-secrets, external-secrets, etc.).
+    existingSecret:
+      # -- Name of the existing Secret. When set, the chart skips creating its own and consumers reference this name.
+      name: ""
+      # -- Optional override of the data key. Defaults to `EAGER_API_KEY`.
+      key: ""
+    # -- Activate the consumer secretKeyRef wiring. When false, downstream env vars and references are omitted entirely.
+    enabled: false
+    # -- Inline API key value. Used only when `existingSecret.name` is empty. Chart creates `<release>-eager-api-key`.
+    value: ""
+
+  # -- Eager-execution client credentials (UNION_CLIENT_ID / UNION_CLIENT_SECRET). Kept separate from `admin` client creds.
+  eagerClientCreds:
+    existingSecret:
+      # -- Name of the existing Secret. When set, the chart skips creating its own.
+      name: ""
+      # -- Optional override of the client-id data key. Defaults to `client_id`.
+      clientIdKey: ""
+      # -- Optional override of the client-secret data key. Defaults to `client_secret`.
+      clientSecretKey: ""
+    enabled: false
+    # -- Inline client id. Used only when `existingSecret.name` is empty.
+    clientId: ""
+    # -- Inline client secret. Used only when `existingSecret.name` is empty.
+    clientSecret: ""
+
+  # -- Image builder registry credentials. Split into push (buildkit) and pull (task pods, replaces depot-token).
+  imageBuilder:
+    push:
+      existingSecret:
+        # -- Name of the existing Secret. When set, the chart skips creating its own.
+        name: ""
+      enabled: false
+      # -- Inline opaque KV map. Each key becomes a Secret data key. Used only when `existingSecret.name` is empty.
+      values: {}
+    pull:
+      existingSecret:
+        # -- Name of the existing Secret. When set, the chart skips creating its own.
+        name: ""
+      enabled: false
+      # -- Inline dockerconfigjson value (full JSON string). Used only when `existingSecret.name` is empty.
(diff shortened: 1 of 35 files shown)


charts/dataplane/values.yaml
```diff
@@ -1875,18 +1875,115 @@ proxy:
...
+    # -- DEPRECATED. Set `existingSecret.name` instead. When `create: false` is left in place the
+    # -- chart treats it as referencing the historical name `union-secret-auth`.
     create: true
     # -- The client secret used to authenticate to the control plane.  This will be provided by Union.
     clientSecret: ""
     # -- The client id used to authenticate to the control plane.  This will be provided by Union.
     clientId: dataplane-operator
 
+  # -- Eager-execution API key (DP → CP callback). Replaces post-install `flyte create secret EAGER_API_KEY`.
+  eagerApiKey:
```
mhotan, line 1926: I am probably going to remove `eagerApiKey` and preserve `eagerClientCreds`.

mhotan:
Superseding this PR. After design review, the direction is shifting away from chart-managed secret mirroring / secrets-by-reference for infra credentials.

The chosen path is **operator self-bootstrap of the eager API key** (config-automation only), reusing the already-merged operator bootstrap: the operator mints/refreshes the eager key and writes the secret, fed by a secret-by-reference seeded into the controlplane chart. Net-new chart surface is a small controlplane values change — no dataplane mirror, no controller-side mount.

Closing in favor of a fresh branch off `main`.

PR #435, opened 2026-06-06
dataplane: render operator connection_config ConfigMap for self-reporting on UpdateStatus

mhotan:
## Summary

When `updateStatus.connectionConfig.host` is set on the dataplane chart, render a ConfigMap with the JSON the operator self-reports on every `UpdateStatus`, and mount it into the operator pod. Companion to controlplane changes in unionai/cloud#16362.

## Changes

- `charts/dataplane/templates/_connection.tpl` — new `dataplane.dp.host` and `dataplane.dp.endpoint` helpers, symmetric with the existing `dataplane.cp.*` helpers.
- `charts/dataplane/templates/operator/configmap-connection.yaml` (new) — renders a `{fullname}-connection` ConfigMap carrying `connection_config.json` with `{ endpoint, insecure, insecureSkipVerify }`. Gated on `host` non-empty.
- `charts/dataplane/templates/operator/deployment.yaml` — mounts the new ConfigMap at `/etc/union/connection-config`, gated the same way.
- `charts/dataplane/templates/operator/configmap.yaml` — sets `operator.connectionConfigPath` in the operator's main config.
- `charts/dataplane/values.yaml` — new `updateStatus.connectionConfig` block. Defaults: `host=""`, `insecure=false`, `insecureSkipVerify=false`.

## Backward compatibility

Pure no-op for envs that don't set `updateStatus.connectionConfig.host`. The gate is on the host helper, so existing installs render no new resources and the controlplane keeps using its legacy admin-set `DataplaneIngressURL`.

Snapshot tests unchanged — no fixture sets the host, so no template output drifts.

## Test plan

- `make generate-expected` — clean (no expected diff).
- `make test` — green.
- `helm template foo charts/dataplane --set updateStatus.connectionConfig.host=dp-1.internal.example.com --set secrets.admin.clientSecret=foo --set secrets.admin.clientId=foo` produces the new ConfigMap, mount, and `operator.connectionConfigPath`.

### DIFF
--- a/charts/controlplane/templates/leasor/configmap.yaml
+++ b/charts/controlplane/templates/leasor/configmap.yaml
@@ -3,12 +3,23 @@
 Bespoke ConfigMap for leasor. Mirrors what the generic services renderer
 (templates/configmap.yaml) would produce — same name, same labels, same
 unionai.configMap-style merge of global .Values.configMap with
-.Values.services.leasor.configMap — and then auto-generates
-queueSync.queues from .Values.global.DATAPLANE_CLUSTER_NAME: one `default`
-queue + one per-cluster queue, both containing that single cluster name. The
-auto-generation is skipped if services.leasor.configMap.leasor.queueSync.queues
-is already set explicitly (full override — use that to enumerate clusters in
-the rare multi-DP case).
+.Values.services.leasor.configMap — then defaults each entry of
+queueSync.queues so callers can write terser overlays.
+
+Per-queue defaults injected when the field is absent from an entry:
+  org:               .Values.global.UNION_ORG  (single-tenant CP convention)
+  name:              "default"
+  priority:          50  (PriorityMedium — see cloud/leasor/config:Priority)
+  fairnessAlgorithm: 1   (FairnessRoundRobin)
+
+The caller still owns `clusters: [...]` (required — there is no sensible
+default) and any other QueueSpec field (project / domain / status /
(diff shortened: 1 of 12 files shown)


charts/controlplane/templates/leasor/configmap.yaml
```diff
@@ -20,20 +31,32 @@ the rare multi-DP case).
...
-{{- $cluster   := .Values.global.DATAPLANE_CLUSTER_NAME | default "" -}}
-{{- $org       := .Values.global.UNION_ORG | default "" -}}
-{{- if and (eq (len $existing) 0) (ne $cluster "") (ne $org "") -}}
-  {{- $queues := list -}}
-  {{- /* default queue pointing at the attached cluster */ -}}
-  {{- $queues = append $queues (dict "org" $org "name" "default"  "priority" 50 "fairnessAlgorithm" 1 "clusters" (list $cluster)) -}}
-  {{- /* per-cluster queue with the same name as the cluster */ -}}
-  {{- $queues = append $queues (dict "org" $org "name" $cluster   "priority" 50 "fairnessAlgorithm" 1 "clusters" (list $cluster)) -}}
-  {{- $_ := set $queueSync "queues" $queues -}}
+{{- $defaultOrg := .Values.global.UNION_ORG | default "" -}}
```
mhotan, line 38: Currently in selfhosted there is no multi-tenancy. It will always be `.Values.global.UNION_ORG`

PR #443, opened 2026-06-12
feat(dataplane): add opt-in FUSE device-plugin DaemonSet

EngHabu:
## What

Adds an opt-in `smarter-device-manager` DaemonSet to the **dataplane** chart that
advertises the host `/dev/fuse` character device as a Kubernetes extended resource
(`smarter-devices/fuse`) via the device-plugin API.

An unprivileged task pod that requests the resource gets `/dev/fuse` injected into
its **devices-cgroup allowlist** by kubelet — which is the one thing a `hostPath`
cannot do — letting it perform an in-pod FUSE mount (e.g. JuiceFS volumes) with
only `CAP_SYS_ADMIN`. No privileged container, no `/dev/fuse` hostPath.

Privilege is confined to the DaemonSet; the workloads it serves stay unprivileged.

## Why

This is the cluster-side half of unprivileged in-pod FUSE. Validated end-to-end on
a k3s cluster: an unprivileged pod requesting `smarter-devices/fuse` + `CAP_SYS_ADMIN`
opened `/dev/fuse` and completed a JuiceFS mount, where the same pod with a `/dev/fuse`
hostPath instead got `EPERM` from the devices cgroup.

## Details

- New templates under `charts/dataplane/templates/fuse-device-plugin/` (daemonset + configmap)
- `fuseDevicePlugin.*` values block, **disabled by default** (`fuseDevicePlugin.enabled: false`)
- `examples/values-fuse-device-plugin.yaml` showing how to enable + a workload's request
- Default-off ⇒ no change to existing rendered output / snapshots.

Companion PR adds the same to the `unionai/cloud` union-operator chart (BYOC).


jeevb (approved):

EngHabu:
@mhotan @davidmirror-ops do you think it's ok to enable the DS by default (given the permissions it requests)?

davidmirror-ops:
> @mhotan @davidmirror-ops do you think it's ok to enable the DS by default (given the permissions it requests)?

Is this for Volumes by any chance? In any case, I don't think we should, especially in light of `low_privilege`, enable by default a container with `CAP_SYS_ADMIN` on every node. It should be opt-in

xjerod (approved):

PR #459, opened 2026-06-30
envoy: fix intracluster CP↔CP routing

chloe-kx:
## Summary
- **Gateway HTTPS/443 intracluster listener**: switch from HTTP/80 to HTTPS/443 with TLS termination for the intracluster listener, matching the public listener.
- **ClientTrafficPolicy h2 ALPN**: add `tls.alpnProtocols: [h2, http/1.1]` — required for grpc-go v1.67+ which enforces ALPN during TLS handshake.
- **`controlPlaneLibrary.ingressFqdn` helper**: named template in `_helpers.tpl` that returns the cluster-local FQDN of the active ingress controller (envoy or nginx) based on `INGRESS_PROVIDER`.
- **`rootTenantURLPattern`**: use the helper instead of hardcoded nginx FQDN, so CP↔CP gRPC routing works for both envoy and nginx.
- **Gateway `https-local` listener + GRPCRoute hostnames**: envoy Gateway uses SNI-based listener matching, so the cluster-local FQDN (`envoy-controlplane.envoy-gateway.svc.cluster.local`) needs its own listener and GRPCRoute hostname entries for CP↔CP traffic to route correctly.
- **GRPCRoute missing services**: add `cloudidl.queue.QueueService`, `cloudidl.queue.InternalQueueService` to cluster rule and `flyteidl2.settings.SettingsService` to organizations rule (parity with nginx ingress).
- **envoy-gateway-config namespace fix**: use `{{ .Release.Namespace }}` instead of hardcoded `envoy-gateway-system` in envoyproxy.yaml, gatewayclass.yaml, and redis.yaml.
- **redis.yaml nil-safe guard**: prevent nil pointer crash when `rateLimit` config block is absent.

## Context
Envoy selfhosted intracluster (DP→CP and CP↔CP) previously used plaintext HTTP on port 80. This switches to HTTPS/443 with `insecureSkipVerify=true`, matching the nginx pattern. All changes are gated behind `INGRESS_PROVIDER=envoy` and are backward compatible with nginx deployments.

## Test plan
- [x] Tested on `chloe-selfhosted` staging cluster
- [x] DP services (executor, union-operator) connect to CP via private DNS on port 443
- [x] CP↔CP routing (executions → flyteadmin) works via envoy cluster-local FQDN
- [x] gRPC ALPN h2 negotiation succeeds with grpc-go v1.67+
- [x] 10 concurrent task pods connect successfully (3 envoy replicas)
- [ ] Verify nginx-only deployments are unaffected

### DIFF
--- a/charts/controlplane/templates/common/_backendtrafficpolicy.yaml
+++ b/charts/controlplane/templates/common/_backendtrafficpolicy.yaml
@@ -1,7 +1,7 @@
 {{- define "control-plane-library.backendtrafficpolicy" }}
 # BackendTrafficPolicy — configures Envoy→gRPC backend connection settings.
 # Replaces nginx grpc_connect_timeout, grpc_read_timeout, grpc_send_timeout.
-# Two policies (one per GRPCRoute) so h2c and timeouts are scoped to gRPC traffic only.
+# Three policies (one per GRPCRoute) so h2c and timeouts are scoped to gRPC traffic only.
 #
 # requestTimeout applies to unary calls; maxStreamDuration applies to streaming calls.
 # "0s" for maxStreamDuration means no limit (equivalent to grpc_read_timeout 604800s on streaming routes).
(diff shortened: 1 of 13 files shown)


charts/controlplane/templates/common/_backendtrafficpolicy.yaml
```diff
@@ -78,6 +78,41 @@ spec:
...
+    tcp:
+      connectTimeout: "1200s"  # grpc_connect_timeout 1200s
+    http:
+      requestTimeout: "1200s"  # grpc_read_timeout 1200s (unary calls)
+      maxStreamDuration: "0s"  # no limit for streaming
+  tcpKeepalive:
+    probes: 9
+    idleTime: "15s"
+    interval: "15s"
+  http2: {}
```
laurabarton, line 102: I think if we end up upgrading to v1.8.0 you can also add in the http2 keepAlive configuraiton.

charts/controlplane/templates/common/_backendtrafficpolicy.yaml
```diff
@@ -78,6 +78,41 @@ spec:
...
+  rateLimit:
+    type: Global
+    global:
+      rules:
+        - clientSelectors:
+            - sourceCIDR:
+                type: Distinct
+                value: "0.0.0.0/0"
+          limit:
+            requests: {{ .Values.envoyGateway.rateLimit.requestsPerUnit | default 100 }}
```
laurabarton, line 113: We'll definitely want to bump this up. We are using 5000 as a default now

charts/controlplane/templates/common/_clienttrafficpolicy.yaml
```diff
@@ -16,6 +16,10 @@ spec:
     http:
       requestReceivedTimeout: "0s"  # client_header_timeout 604800
       streamIdleTimeout: "0s"       # client_body_timeout 604800
+  tls:
+    alpnProtocols:
```
laurabarton, line 20: This is new and is not on the manage cluster setup, might be a good change to test out there as well.

laurabarton (approved):

PR #469, opened 2026-07-08
Add zero-trust overrides

davidmirror-ops:


jeevb (approved):

charts/dataplane/values.zero-trust.yaml
```diff
@@ -0,0 +1,21 @@
+# Opt-in zero-trust mode (https://www.union.ai/docs/v2/union/security/).
```
katrogan, line 1: nit: can we add a comment that this is recommended (required even) to be enabled

katrogan (approved):

charts/dataplane/values.zero-trust.yaml
davidmirror-ops: just added, let me know if that works

PR #474, opened 2026-07-10
[Leaseworker] Add priority class

wild-endeavor:
## Overview

Backports unionai/cloud#16230 to this chart: leaseworker here still runs at default pod
priority, so a replacement pod can sit Pending behind node scale-up — and since it's a
Recreate singleton, that means no leases flow on the dataplane until it schedules.

Adds `leaseworker.priorityClassName`, default `system-cluster-critical` (matching
flytepropeller in this chart). Set it to `""` to render nothing.

## Test Plan

`make generate-expected` + `make test` pass. Every dataplane snapshot changes by exactly
the one expected line.


charts/dataplane/values.yaml
```diff
@@ -1093,6 +1093,8 @@ leaseworker:
   additionalVolumeMounts: []
   # -- Appends additional environment variables to the leaseworker container's spec.
   podEnv: []
+  # -- Sets priorityClassName for the leaseworker pod.
+  priorityClassName: "system-cluster-critical"
```
mhotan, line 1097: Looks like it is shipped with Kubernetes. https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#how-to-use-priority-and-preemption. So guaranteed to exist.

mhotan (approved):

pvditt (approved):

SVilgelm (approved):

PR #478, opened 2026-07-15
dataplane: custom PriorityClass for flytepropeller (enabled + low_privilege aware)

mhotan:
## Summary

Follow-up to #477 (leaseworker). flytepropeller still defaulted `priorityClassName: system-cluster-critical` — a **reserved Kubernetes system class** that GKE's `gcp-critical-pods` ResourceQuota restricts to `kube-system`. It's dormant today because `flytepropeller.enabled` defaults `false`, but the moment propeller is enabled on GKE its pod would `FailedCreate` exactly like the leaseworker did. This applies the same chart-managed custom-PriorityClass pattern to flytepropeller.

## Changes

- `flytepropeller.priorityClassName` default → `union-flytepropeller` (custom, non-system).
- `flytepropeller.priorityClass.{create,value}` — the chart renders `templates/propeller/priorityclass.yaml`, gated on **`flytepropeller.enabled` AND `not low_privilege`**. Set `create: false` to bring your own class.
- The deployment's `priorityClassName` reference is BYO/`low_privilege`-aware (matching leaseworker): creation is gated on `low_privilege`, but an externally-managed class is still referenced.

## `enabled` + `low_privilege` behavior

| `enabled` | `create` | `low_privilege` | Creates class | Pod references it |
|---|---|---|---|---|
| false | — | — | ❌ | ❌ (deployment not rendered) |
| true | true | false | ✅ | ✅ |
| true | true | true | ❌ | ❌ (no dangling reference) |
| true | false (BYO) | either | ❌ | ✅ (externally managed) |

## Test Plan

- New `tests/values/dataplane.flytepropeller.yaml` (`flytepropeller.enabled: true`, `low_privilege: false`) — snapshot renders the `union-flytepropeller` PriorityClass + the deployment reference; no `system-cluster-critical` reference remains. No existing snapshot moves (flytepropeller is disabled in every other fixture — which is itself the "respects `enabled`" behavior).
- Verified the full matrix via `helm template --set` (enabled/disabled × create × low_privilege × BYO).

## Rollback

`git revert` — flytepropeller returns to referencing `system-cluster-critical` (unchanged latent behavior; still gated behind `enabled: false`).

### DIFF
--- a/charts/dataplane/templates/propeller/deployment.yaml
+++ b/charts/dataplane/templates/propeller/deployment.yaml
@@ -26,7 +26,7 @@ spec:
       {{- with .Values.flytepropeller.securityContext }}
       securityContext: {{ tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
-      {{- if and .Values.flytepropeller.priorityClassName (not .Values.low_privilege) }}
+      {{- if and .Values.flytepropeller.priorityClassName (or (not .Values.flytepropeller.priorityClass.create) (not .Values.low_privilege)) }}
       priorityClassName: {{ .Values.flytepropeller.priorityClassName }}
       {{- end }}
       containers:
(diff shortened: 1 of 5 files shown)


wild-endeavor (approved):
can we just retire flyte propeller soon? :(

mhotan:
> can we just retire flyte propeller soon? :(

Hmm that is a good question. This probably tied to v1 retirement, right?

PR #486, opened 2026-07-21
Use cluster name when minting eager api key

katrogan:
Not to be merged until updated with latest cloud image

- `main` 
  - **Use cluster name when minting eager api key** :point\_left:
    - \#488

### DIFF
--- a/charts/dataplane/Chart.yaml
+++ b/charts/dataplane/Chart.yaml
@@ -3,7 +3,7 @@ name: dataplane
 description: Deploys the Union dataplane components to onboard a kubernetes cluster to the Union Cloud.
 type: application
 icon: https://i.ibb.co/JxfDQsL/Union-Symbol-yellow-2.png
-version: 2026.7.2
+version: 2026.7.1
 appVersion: 2026.7.0
 kubeVersion: '>= 1.28.0-0'
 dependencies:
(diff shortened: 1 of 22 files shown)


charts/dataplane/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: dataplane
 description: Deploys the Union dataplane components to onboard a kubernetes cluster to the Union Cloud.
 type: application
 icon: https://i.ibb.co/JxfDQsL/Union-Symbol-yellow-2.png
-version: 2026.7.1
+version: 2026.7.2
```
mhotan: nit: it would be great to do a release later. Collect a set of changes, update PR descripton with release notes, review, and update release notes later.

mhotan (approved):

mhotan:
Validated end-to-end on a two-dataplane, single-org selfmanaged staging env (one AWS dataplane + one GCP dataplane, separate secret stores, against a managed control plane) — the multi-dataplane rotation-war scenario. Each dataplane's operator minted a distinct per-cluster OAuth app (`…-EAGER_API_KEY-<cluster>`), the pod webhook resolved the fixed `EAGER_API_KEY` to the cluster-scoped secret, and both dataplanes' V2 functional tests passed simultaneously with no rotation war. The chart change slots in cleanly.

mhotan:
@katrogan merged https://github.com/unionai/helm-charts/pull/488 to revert the release along with code change. We can release at a later time. Considering this depends on a Cloud change we should stabilize, publish new images, and release the chart at a later time.

mhotan:
Marking as ready for review as well.

PR #487, opened 2026-07-21
dataplane: per-DP Eager API key control-plane OAuth app name

mhotan:
## Overview

The operator self-bootstraps an Eager API key by minting a control-plane OAuth app named after a fixed key. When one organization runs **two data planes**, both mint/rotate the *same* org-scoped app, invalidating each other's stored key.

This adds an `eagerApiKeyName` value — `EAGER_API_KEY`, plus a `_<CLUSTER_NAME>` suffix when operator apiKey bootstrapping is enabled (`config.operator.apiKey.enabled`) — and wires it to `config.operator.apiKey.secretName`, the name of the control-plane OAuth app the operator mints. Co-located data planes then get **distinct** apps instead of rotating a shared one.

**Scope of the name.** `eagerApiKeyName` names *only* the control-plane OAuth app. Task pods resolve the Eager key under the fixed `EAGER_API_KEY` (flyteadmin's hard-coded lookup), so every `unionAuth.secretName` stays `EAGER_API_KEY` — making that per-DP would leave task pods unable to find the key. The default (bootstrapping off, or no cluster name) is the canonical `EAGER_API_KEY`, so single-DP deployments are unchanged.

## Testing

- `make generate-expected` + `make test` green. The only rendered change is a quoting normalization of the existing `unionAuth.secretName: EAGER_API_KEY` across snapshots; the value is unchanged.
- Verified live on a two-data-plane, single-org environment: both data planes run workflows simultaneously, each minting its own OAuth app, with task pods authenticating via client credentials.

Requires the companion operator change that makes the CreateKey key configurable (defaults to `EAGER_API_KEY`).

### DIFF
--- a/charts/dataplane/values.aws.yaml
+++ b/charts/dataplane/values.aws.yaml
@@ -164,20 +164,6 @@ config:
       enabled: false
 
 # ----------------------------------------------------------------------------
-# Executor Configuration
-# ----------------------------------------------------------------------------
-# Configure how workflow tasks are executed.
-
-executor:
-  config:
-    unionAuth:
-      # --- API key injection ---
-      # Injects the EAGER_API_KEY secret into task pods
-      # for authenticated eager-mode execution.
-      injectSecret: true
-      secretName: "EAGER_API_KEY"
-
-# ----------------------------------------------------------------------------
 # Ingress Configuration
 # ----------------------------------------------------------------------------
 # Configure ingress for external access to dataplane services.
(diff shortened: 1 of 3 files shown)


mhotan:
**Tabling — preserving the branch as reference.**

Defaulting `apiKey.secretName` to a fixed name only holds when data planes don't share a secret store. Converging on the cluster-scoped eager-key approach in #486, which demarcates the stored secret by cluster (with the corresponding operator/IDL changes). This chart-side default isn't needed under that approach.

mhotan:
Closing — superseded by the cluster-scoped eager-key approach in #486, which handles shared secret stores that this chart-side default doesn't. Branch preserved for reference.

PR #491, opened 2026-07-23
controlplane: apiKeyOverrides as a list for per-cluster seeded API keys

mhotan:
## Overview

`services.identity.apiKeyOverrides` changes from a **map keyed by system-key name** to a **list** of `{key, clusterName?, existingSecret}` entries, so a control plane can seed a distinct OAuth client per data plane for the same system key (e.g. a per-cluster `EAGER_API_KEY`). An entry without `clusterName` is the cluster-nameless default; a cluster-scoped entry takes precedence for that data plane. Mount paths are now per-entry (`/etc/secrets/apikey/<KEY>` or `<KEY>-<clusterName>`).

This is the chart half of per-dataplane operator + eager OAuth clients; the cloud terraform that emits the per-cluster list is unionai/cloud#17168. The identity backend already keys `apiKeyOverride` lookup on org+key+clusterName with a nameless fallback.

## Breaking change / migration

Values-shape change (map to list) — documented in `charts/MIGRATION.md`. Chart `version` bumped. Env overlays that still carry the map shape must regenerate (terraform re-apply for selfmanaged) in lockstep with bumping to this chart version.

## Test Plan

- `helm lint` + `helm template` render clean; snapshot `tests/generated/controlplane.custom-oidc.yaml` regenerated and now exercises two entries (a nameless default + a cluster-scoped `dp-1`), producing distinct volumes, mount dirs, and referenced secrets.
- Rendered end-to-end via the cloud terraform on `mike-apple-aws` (per-cluster `apiKeyOverrides` for dp-1/dp-2) and deployed via `internal-selfmanaged-staging-sync` #1046.

## Rollout

Per-env, gated by each env's chart-revision pin. `mike-apple-aws` perennial bumped to this branch; other envs adopt on their next bump (with a coordinated terraform re-apply).

## Rollback

`git revert`; envs pinned to the prior chart revision are unaffected until they re-pin.

### DIFF
--- a/charts/MIGRATION.md
+++ b/charts/MIGRATION.md
@@ -4,6 +4,71 @@ Tracking the migration story for the helm-charts cloud overlays. New entries at 
 
 ---
 
+## controlplane — `services.identity.apiKeyOverrides` map → list (per-cluster seeded API keys)
+
+### What changed
+
+`services.identity.apiKeyOverrides` changes from a **map keyed by system-key name** to a
+**list** of entries, each identified by `key` + optional `clusterName`. This lets one control
+plane seed a *different* OAuth client per dataplane for the same system key — the map shape
+allowed only one entry per key.
+
+Before (map):
+
+```yaml
+services:
+  identity:
+    apiKeyOverrides:
+      EAGER_API_KEY:
+        existingSecret:
+          name: eager-client-creds
+```
+
+After (list):
+
+```yaml
+services:
+  identity:
+    apiKeyOverrides:
+      - key: EAGER_API_KEY            # cluster-nameless default (serves any dataplane)
+        existingSecret:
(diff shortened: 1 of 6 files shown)


charts/controlplane/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: controlplane
 description: Deploys the Union controlplane components to onboard a kubernetes cluster to the Union Cloud
 type: application
 icon: https://i.ibb.co/JxfDQsL/Union-Symbol-yellow-2.png
-version: 2026.7.1
+version: 2026.7.2
```
mhotan: Don't update version, leave for a later specific release PR.

xjerod (approved):

PR #497, opened 2026-07-27
dataplane: make cloud overlays cloud-only (hoist topology + platform defaults to base)

mhotan:
## Overview

Reduces `charts/dataplane/values.{aws,gcp,azure}.yaml` to **cloud-only** config — provider, object storage, IAM / Workload-Identity annotations, region + task-log wiring, and the cloud globals a user fills in. Everything else (deployment-topology and Union-platform config) that the overlays previously restated per cloud — and, in places, silently lost to the terraform config-subtree shallow-merge, so the same env type rendered different values per cloud — moves to base `values.yaml` chart defaults, set per-topology by the `union_extension` terraform (cloud #17241). Each overlay now opens with a PURPOSE header documenting the cloud-only contract. See `charts/dataplane/RELEASE.md` (Unreleased) for the full list.

**Hoisted to base / removed from overlays:**

- **admin `clientId`** derives from the single `global.AUTH_CLIENT_ID` (dropped 9 overlay restatements); base default is empty, not the `dataplane-operator` placeholder.
- **cross-cloud monitoring + storage defaults** — `storage.enableMultiContainer: true`; `monitoring.kube{ControllerManager,Scheduler,Etcd,Proxy}.enabled: false` + `defaultRules.rules.*: false` (a managed/co-located CP owns those k8s control-plane components; a dataplane can't scrape them and would trip false-positive TargetDown alerts).
- **auth enablement** — `config.union.auth.enable`, `clusterresourcesync…auth.enable`+scopes, and catalog-cache `use-admin-auth` default **enabled** in base; terraform disables only for a selfhosted CP with no IdP.
- **ingress / ingress-nginx** — config in base (`enabled: false`); terraform enables both only for a co-located CP (selfhosted).
- **secrets.admin** / **serving.auth** / **dcgm-exporter.enabled** / **controlplaneNamespace** / task-pod `default-env-vars` — base defaults; per-cloud overrides dropped (the leaseworker + executor already inject the `_U_*` task-pod vars from `config.union.connection`; dcgm keeps only its cloud-specific node affinity).
- **dead keys** — `prometheus.prometheusOperator`, empty `flytepropeller` stubs, `billableUsageCollector.enabled`, `singleTenantOrgID`, base restatements.

**Out of scope (by design):** the empty `global.*` placeholders stay — they intentionally document the per-cloud set of globals a user must fill in.

## Testing

- `make helm-test` green — the base hoists are render-neutral across the snapshot fixtures (zero snapshot drift).
- Base+overlay render is byte-identical where a real deployment sets the value; where a base default now applies (e.g. `use-admin-auth`, `serving.auth`) the value matches what the overlay used to assert.
- Cross-topology deploy validation via the mike staging envs tracking `mike/head` (managed-CP aws/gcp/azure + selfhosted-intracluster gcp) through `internal-selfmanaged-staging-sync-all` — functional tests green.

## Related

Terraform half: unionai/cloud #17241 (gates the topology-specific settings). Deploy in lockstep.

- `main` 
  - **dataplane: make cloud overlays cloud-only (hoist topology + platform defaults to base)** :point\_left:
    - \#498
      - \#499


charts/dataplane/values.yaml
```diff
@@ -2448,16 +2467,29 @@ monitoring:
...
     create: true
+    rules:
+      kubeControllerManager: false
+      kubeSchedulerAlerting: false
+      kubeSchedulerRecording: false
+      kubeProxy: false
+      etcd: false
   kubeApiServer:
     enabled: true
+  # Assumes a managed Kubernetes control plane (EKS/GKE/AKS) or the Union managed
```
davidmirror-ops: So these new defaults are useful mostly for self hosted?

mhotan: No the default are actually oriented toward selfmanaged. Of the course of the past few weeks I made it so the divergence between selfhosted CP and Union managed CP is minimal

mhotan: This comment is in reference to the Kubernetes. Not Union Control Plane. Ill update

mhotan: We should update the comment to be enable if you are confident that your Kubernetes provider allows access to these metrics

mhotan:
Reworded in the latest commit — the comment now frames it around the Kubernetes provider (EKS/GKE/AKS) not exposing control-plane component metrics, decoupled from the Union CP / selfhosted-vs-managed distinction, with an explicit 'enable if your Kubernetes provider allows access' note.

xjerod (approved):

PR #499, opened 2026-07-28
dataplane: consolidate control-plane host to a single CONTROLPLANE_HOST

mhotan:
## Overview

The DP→CP destination was four overlapping globals (`CONTROLPLANE_HOST`, `UNION_CONTROL_PLANE_HOST`, `CONTROLPLANE_GRPC_ENDPOINT`, `QUEUE_GRPC_ENDPOINT`). This collapses them to **one host**.

- **`CONTROLPLANE_HOST`** is the single host global, moved to the **top** of the globals ("configure first"). `dataplane.cp.endpoint` derives `dns:///<host>` from it.
- **Removed `CONTROLPLANE_GRPC_ENDPOINT`** — always derived (`dns:///<host>`). The `:443` port is omitted; grpc-go's DNS resolver and Connect/https both default to 443.
- **Removed `QUEUE_GRPC_ENDPOINT` + the `dataplane.cp.queueEndpoint` helper** — the task-pod endpoint is injected by the leaseworker/executor from `config.union.connection`; there is no separate queue global. Authless / direct-service routing (skip nginx + OAuth, dial the in-cluster Service on `:80`) is configured explicitly via `config.union.auth` + `config.union.connection`, documented in the new **`examples/values.authless.yaml`**.
- **`UNION_CONTROL_PLANE_HOST` and top-level `.Values.host`** are marked **DEPRECATED** — still honored as fallbacks (precedence `CONTROLPLANE_HOST` > `.Values.host` > `UNION_CONTROL_PLANE_HOST`), migrate to `CONTROLPLANE_HOST`. No terraform env sets `.Values.host`.

Paired terraform (unionai/cloud): `union_extension` stops emitting the three retired globals; all four mike staging envs regenerated.

## Testing

`make helm-test` green. Snapshot diff is exactly the `:443` drop at every endpoint site (293) + derived configChecksums — nothing else. Deployed to all four mike staging envs (managed-CP aws/gcp/azure + selfhosted-intracluster/split gcp + aws-CP) via `internal-selfmanaged-staging-sync-all` to confirm the task-pod SDK resolves 443 without the explicit port.

## Migration

Any env whose generated values still set the retired globals or reference `dataplane.cp.queueEndpoint` in task-pod `default-env-vars` must be regenerated before adopting (the helper is removed). See RELEASE.md.

## Stack

Stacked on #498 → #497. Part of the selfmanaged overlay/host cleanup.

- `main` 
  - \#498
    - **dataplane: consolidate control-plane host to a single CONTROLPLANE\_HOST** :point\_left:

### DIFF
--- a/charts/dataplane/examples/values.authless.yaml
+++ b/charts/dataplane/examples/values.authless.yaml
@@ -1,0 +1,60 @@
+# ============================================================================
+# Authless dataplane example
+# ============================================================================
+# Shared-cluster deployment that SKIPS auth and routes DP -> CP traffic directly
+# to in-cluster Kubernetes Services, bypassing the control-plane nginx ingress
+# and OAuth entirely. This is the historical "shared cluster, skip auth, route
+# to k8s services" topology.
+#
+# Only safe when the control plane and data plane share a cluster and the CP
+# Services are not reachable from outside it — the control plane trusts these
+# in-cluster requests via identity claims instead of an OAuth token.
+#
+# How it works:
+#   - The single host global CONTROLPLANE_HOST still names the CP for the
+#     serving / app-callback URLs and cloudHostName.
+#   - The DP -> CP gRPC connection is overridden to the in-cluster Service on
+#     :80 with insecure transport. The `…svc.cluster.local:80` + insecure
+#     combination is exactly what selects the control plane's trusted-identity
+#     (no-OAuth) code path — so no token is required.
+#   - Auth is turned off explicitly via config.union.auth.enable (and the
+#     sibling toggles), NOT derived from the host. These are the individual
+#     auth toggles the interface exposes.
+#
+# Replace `union-cp` with the actual CP release namespace, and the Service names
+# with the CP's exposed gRPC Services, if they differ in your deployment.
+#
+# Apply on top of values.yaml + a cloud overlay:
+#   helm template charts/dataplane \
+#     -f charts/dataplane/values.yaml \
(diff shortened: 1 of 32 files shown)


charts/dataplane/examples/values.authless.yaml
```diff
@@ -0,0 +1,60 @@
...
+    auth:
+      enable: false
+    # Dial the control plane directly in-cluster, plaintext. The
+    # svc.cluster.local:80 authority + insecure activates the CP's
+    # trusted-in-cluster-identity path (no OAuth, no nginx).
+    connection:
+      host: "dns:///union.union-cp.svc.cluster.local:80"
+      insecure: true
+
+# Run FULLY authless by mirroring the same connection + disabling auth on every
```
mhotan: We should show each config block we need to disable auth for.

mhotan: Done in 77e804d0 — the example now spells out each DP→CP client block (union.connection, admin, catalog-cache, authorizer, clusterresourcesync, + optional serving/gateway app auth) as explicit YAML pointing at the CP's in-cluster gRPC Services on :80 + insecure.

mhotan:
Addressed in 77e804d0 — the authless example now spells out each DP→CP client block that needs auth disabled (union.connection, admin, catalog-cache, authorizer, clusterresourcesync, plus optional serving/gateway app auth).

wild-endeavor (approved):

PR #504, opened 2026-07-31
controlplane: default rootTenantURLPattern to global.UNION_HOST

mhotan:
## Overview

`connection.rootTenantURLPattern` is the host identity embeds in minted api-keys (e.g. `EAGER_API_KEY` / operator eager-key bootstrap). The chart defaulted it to the cluster-local ingress svc FQDN (`controlPlaneLibrary.ingressFqdn`), which only resolves **intracluster** — an operator in a separate DP cluster can't reach it.

Default it instead to the publicly-resolvable control-plane host `dns:///{{ .Values.global.UNION_HOST }}`, which works for both topologies. Envs that want optimized in-cluster / internal-VPC routing override it — the selfmanaged `union_extension` module derives the topology-aware host (svc FQDN intracluster, routable host separate-cluster) and sets it in the generated values (unionai/cloud #17367).

Regenerated snapshots (`make generate-expected`).

## Test Plan

- `make generate-expected` + `make test` — snapshots updated to render `dns:///{UNION_HOST}` instead of the svc FQDN.

## Rollback

`git revert`.

### DIFF
--- a/charts/controlplane/values.yaml
+++ b/charts/controlplane/values.yaml
@@ -500,8 +500,10 @@ configMap:
   connection:
     environment: "staging"
     region: ""  # Cloud-specific: set in overlay (e.g. AWS_REGION)
-    # Use controlplane ingress controller for gRPC routing.
-    rootTenantURLPattern: 'dns:///{{ include "controlPlaneLibrary.ingressFqdn" . }}'
+    # Operator eager-key (CreateAPI) bootstrap dial. Default to the publicly
+    # resolvable control-plane host; override per-env (global.CONTROLPLANE_HOST /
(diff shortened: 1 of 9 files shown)


mhotan:
Closing — rootTenantURLPattern is the **cluster-local CP↔CP gRPC host** (see _helpers.tpl:46 + the Go consumers in organizations/artifacts/cluster/identity), not a public/SDK host. Defaulting it to global.UNION_HOST would route CP↔CP gRPC through the public LB. The chart default should remain the intra-cluster ingressFqdn; the cloud union_extension derives the topology-aware internal host (unionai/cloud #17367). A follow-up will have the controlplane module expose an explicit internal_host output.

PR #508, opened 2026-08-05
ci: trigger cloud canary deploys on chart merges to main

mhotan:
## Overview

On merge to `main`, poke the Union cloud Buildkite pipelines that keep the canary self-managed fleet in sync with this repo, routed by which chart files changed. Implemented as a `push`-to-`main` GitHub Action (`.github/workflows/trigger-union-internal-checks.yaml`) that calls the Buildkite REST API via a small helper (`.github/scripts/trigger-buildkite-build.sh`).

## Routing

| Changed | Pipeline triggered |
| --- | --- |
| `charts/*/values.aws.yaml`, `values.gcp.yaml`, `values.azure.yaml` | `tf-fanout-canary-deploy` |
| `charts/*/values.yaml`, `charts/*/templates/**`, `charts/*/Chart.yaml` | `internal-selfmanaged-canary-sync-all` |

A merge touching both classes fires both. Scope is broad by design (all canary self-managed): `DIR_LIST` / `ORGS_LIST` are left unset, which both pipelines treat as "the whole canary tier". `SKIP_INPUT=TRUE` is sent to the fan-out so it auto-applies. Each build carries a `helm-charts@<sha> by <actor>` provenance stamp.

## Auth

The Action authenticates with a `BUILDKITE_API_TOKEN` repo secret — a Buildkite API token scoped to the `unionai` organization with the `write_builds` REST scope (nothing else). It must be provisioned once in repo settings:

```
gh secret set BUILDKITE_API_TOKEN --repo unionai/helm-charts --body '<token>'
```

## Test Plan

- `yaml.safe_load` on the workflow — parses.
- `bash -n` on the helper script — clean.
- Router logic exercised against a sample changed-file list: overlays route to the fan-out, base `values.yaml` / nested `templates/**` / `Chart.yaml` route to sync-all, and non-chart files (e.g. `values.k3d.yaml`, `README.md`) are correctly ignored.
- End-to-end trigger fires only once the `BUILDKITE_API_TOKEN` secret is provisioned and the target pipelines are live.

### DIFF
--- a/.github/scripts/trigger-buildkite-build.sh
+++ b/.github/scripts/trigger-buildkite-build.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Create a Buildkite build on a cloud pipeline via the REST API.
+#
+# Usage: trigger-buildkite-build.sh <pipeline-slug> [KEY=VALUE ...]
+#
+# Each KEY=VALUE becomes a build env var. TRIGGER_PROVENANCE is always added,
+# stamped as "helm-charts@<sha> by <actor>" so the triggered cloud pipeline
+# (and its PR-back / annotation) can attribute the run to this merge.
+#
+# The build runs cloud's main (commit HEAD, branch main): the target pipelines'
+# repository is unionai/cloud, so they must build cloud, not helm-charts.
+#
+# Requires: BUILDKITE_API_TOKEN in the environment (Buildkite API token with the
+# write_builds scope), plus GITHUB_SHA and GITHUB_ACTOR (set by Actions).
+set -euo pipefail
+
+pipeline="${1:?pipeline slug required}"
+shift || true
+
+org="unionai"
+provenance="helm-charts@${GITHUB_SHA} by ${GITHUB_ACTOR}"
+
+# Build the env object from the KEY=VALUE args plus provenance.
+env_json='{}'
+for kv in "$@"; do
+  key="${kv%%=*}"
+  val="${kv#*=}"
+  env_json="$(jq -c --arg k "$key" --arg v "$val" '. + {($k): $v}' <<<"$env_json")"
+done
+env_json="$(jq -c --arg v "$provenance" '. + {TRIGGER_PROVENANCE: $v}' <<<"$env_json")"
+
+body="$(jq -n \
+  --arg commit "HEAD" \
+  --arg branch "main" \
+  --arg message "$provenance: trigger ${pipeline}" \
+  --argjson env "$env_json" \
+  '{commit: $commit, branch: $branch, message: $message, env: $env}')"
+
+echo "Triggering ${org}/${pipeline} with env: ${env_json}"
+curl --fail --silent --show-error \
+  -X POST \
+  -H "Authorization: Bearer ${BUILDKITE_API_TOKEN}" \
+  "https://api.buildkite.com/v2/organizations/${org}/pipelines/${pipeline}/builds" \
+  -d "$body" \
+  | jq -r '"Created build: \(.web_url)"'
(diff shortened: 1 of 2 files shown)


.github/workflows/trigger-canary-deploy.yaml
```diff
@@ -0,0 +1,85 @@
...
+# API token with the write_builds scope). The target pipelines live in unionai/cloud
+# (infra/terraform/envs/canary/helm-charts-ci); the token's existence is checked there
+# by ci_trigger_token.tf.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'charts/*/values.yaml'
+      - 'charts/*/values.aws.yaml'
```
mhotan: Make this more robust if possible. We should trigger if not values.k3d.yaml but match any values.<cloud>.yaml

.github/workflows/trigger-canary-deploy.yaml
```diff
@@ -0,0 +1,85 @@
+name: trigger-canary-deploy
```
mhotan: Can we rename to "Trigger Union Internal Checks"

.github/workflows/trigger-canary-deploy.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+# On merge to main, poke the cloud (unionai/cloud) Buildkite pipelines that keep
```
mhotan: Remove this long comment, it's not necessary.

.github/workflows/trigger-canary-deploy.yaml
mhotan: Done in `fbfc71ff`. The router now matches any `values.*.yaml` overlay generically, so a new cloud (e.g. `values.oci.yaml`) routes to `tf-fanout-canary-deploy` with no edit here. Non-deployed overlays are special-cased: `values.k3d.yaml` is ignored (local dev overlay, no k3d env in the canary fleet) and base `values.yaml` goes to `internal-selfmanaged-canary-sync-all`. Also added `!charts/*/values.k3d.yaml` to the `paths:` gate. Verified the routing against aws/azure/oci/k3d/base/templates/Chart.yaml/sandbox samples.

.github/workflows/trigger-canary-deploy.yaml
mhotan: Done in `f9b2f313` — `name: Trigger Union Internal Checks`. Left the filename as `trigger-canary-deploy.yaml`; happy to rename the file to match if you'd prefer.

.github/workflows/trigger-canary-deploy.yaml
mhotan: Done in `f9b2f313` — trimmed to a 2-line header (purpose + auth). The `paths:` globs and step names carry the routing detail.

mhotan:
**On `BUILDKITE_API_TOKEN` security (public repo):**

- Encrypted Actions **repo secret**, never committed; the workflow only references `${{ secrets.BUILDKITE_API_TOKEN }}` (passed via env var, never echoed), and GitHub auto-masks it in logs.
- Triggers on **`push: [main]` only** (not `pull_request` / `pull_request_target`), so fork PRs never receive the secret — the classic public-repo exfiltration path is closed. The secret-bearing run only executes on already-merged code.
- **Intentionally not scoped to the `helm-charts-ci` environment**: that env's FAB-430 App protection rule gates on the deployment creator's `all-eng` membership, which would fail post-merge triggers whose author is a non-member or a bot (aviator/mergequeue) — wrong semantics for a post-merge deploy trigger.
- Hardening (`f72140fb`): `permissions: contents: read` (job needs no other `GITHUB_TOKEN` scopes) + `if: github.repository == 'unionai/helm-charts'` (forks no-op).
- Token is least-privilege: org `unionai`, REST scope `write_builds` only (no read/pipeline/agent access). Recommend a rotating bot token with an expiry for steady state.

xjerod (approved):

PR #514, opened 2026-08-07
fix(dataplane): make runtime settings configurable

xjerod:
## Overview

Makes three dataplane runtime settings configurable while retaining their existing defaults:

- `leaseworker.podSecurityContext` now controls the leaseworker pod security context. It defaults to the existing `fsGroup: 1337`.
- `fluentbit.tailDBPath` now controls the Fluent Bit tail input state database path. It defaults to the existing `/var/log/flb_kube.db`.
- `serviceAccount.imagePullSecrets` now controls image pull secrets on the single-namespace `union` task service account. `commonServiceAccount.imagePullSecrets` remains scoped to `union-system`.

A focused `dataplane.runtime-overrides` fixture exercises all three settings with non-default values, including distinct pull secrets for the task and system service accounts.

## Compatibility

Existing rendered output is unchanged with default values. No values migration is required.

## Testing

- `make generate-expected`
- `make test`
  - Helm snapshot tests
  - CRD synchronization checks
  - kubeconform validation
  - image-path validation

The full test suite passed after rebasing onto the current `main` and regenerating snapshots for the current chart and app versions.

## Stack

- `main` 
  - **fix(dataplane): make runtime settings configurable** :point\_left:
    - \#515
      - \#516

### DIFF
--- a/charts/dataplane/templates/common/union-serviceaccount.yaml
+++ b/charts/dataplane/templates/common/union-serviceaccount.yaml
@@ -7,4 +7,8 @@ metadata:
   annotations:
     {{ tpl .Values.userRoleAnnotationKey . }}: {{ tpl .Values.userRoleAnnotationValue . }}
 automountServiceAccountToken: true
+{{- with .Values.serviceAccount.imagePullSecrets }}
+imagePullSecrets:
+{{- tpl (toYaml .) $ | nindent 2 }}
 {{- end }}
+{{- end -}}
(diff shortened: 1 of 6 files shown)


charts/dataplane/templates/common/union-serviceaccount.yaml
```diff
@@ -7,4 +7,8 @@ metadata:
   annotations:
     {{ tpl .Values.userRoleAnnotationKey . }}: {{ tpl .Values.userRoleAnnotationValue . }}
 automountServiceAccountToken: true
-{{- end }}
\ No newline at end of file
+{{- with .Values.commonServiceAccount.imagePullSecrets }}
```
mhotan, line 10: nit: we should be able to use `serviceAccount`. E.g. .Values.serviceAccount.imagePullSecrets. We kind of have the pattern that top level values are in the injection chain. https://github.com/unionai/helm-charts/blob/6e610067576033a09e7cd6e6f57cc4666fbd09bb/charts/dataplane/values.yaml#L69-L85

mhotan (approved):

fiedlerNr9 (approved):

charts/dataplane/templates/common/union-serviceaccount.yaml
xjerod: Updated this to use top-level `serviceAccount.imagePullSecrets` for the single-namespace `union` task service account. `commonServiceAccount.imagePullSecrets` remains scoped to `union-system`, and the runtime fixture uses distinct secrets to cover both paths.

PR #515, opened 2026-08-07
fix(controlplane): harden actions chart rendering

xjerod:
## Overview

Makes sharded Actions deployments more portable and Helm snapshot generation safer.

- Adds `actions.coordination.securityContext` so the coordination init container can run with deployment-specific security settings, including dynamically assigned UIDs.
- Quotes shard-hash labels and selectors so they consistently render as strings.
- Adds focused coverage for overriding the Actions coordination security context.
- Generates snapshots through temporary files and replaces checked-in output only after Helm succeeds.
- Adds a regression test verifying failed renders preserve the existing snapshot and propagate the original exit status.

## Testing

- `make generate-expected`
- `make test`
- `make snapshot-generator-test`
- `bash -n tests/run.sh tests/lib/atomic-render.sh tests/test-atomic-render.sh`
- `git diff --check`

## Stack

- `main` 
  - **fix(controlplane): harden actions chart rendering** :point\_left:
    - \#516

### DIFF
--- a/charts/controlplane/values.yaml
+++ b/charts/controlplane/values.yaml
@@ -1510,6 +1510,13 @@ actions:
     image:
       repository: docker.io/alpine/k8s
       tag: "1.32.3"
+    securityContext:
+      runAsNonRoot: true
+      runAsUser: 65534
+      allowPrivilegeEscalation: false
+      readOnlyRootFilesystem: true
+      capabilities:
+        drop: ["ALL"]
 
   serviceAccount:
     create: true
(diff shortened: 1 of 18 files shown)


charts/controlplane/values.yaml
```diff
@@ -1492,6 +1492,13 @@ actions:
     image:
       repository: docker.io/alpine/k8s
       tag: "1.32.3"
+    securityContext:
```
mhotan, line 1495: 👍

mhotan (approved):

fiedlerNr9 (approved):

PR #516, opened 2026-08-07
feat(dataplane): add OpenShift compatibility support

xjerod:
## Overview

Adds an opt-in OpenShift compatibility path for the dataplane chart.

- Adds `values.openshift.yaml` with rootless BuildKit defaults, a dedicated service account, `hostUsers: false`, `/tmp` task working directories, and ephemeral Fluent Bit state.
- Adds configurable SecurityContextConstraints and namespaced RBAC for BuildKit and the Kourier gateway.
- Supports either creating chart-managed SCCs or binding existing SCCs.
- Makes the KnativeServing Helm hook annotations configurable.
- Adds a validation script for BuildKit rollout, SCC admission and access, worker readiness, and an optional build-and-push smoke test.
- Adds snapshot coverage for both chart-created and pre-existing SCC configurations.
- Documents how to layer and validate the OpenShift values overlay.

The OpenShift behavior is opt-in. Regenerating snapshots did not change existing fixtures.

## Testing

- `helm lint charts/dataplane` — passed
- `make generate-expected` — passed
- `make test` — passed
- `bash -n scripts/validate_rootless_buildkit.sh` — passed
- `git diff --check` — passed
- Verified focused renders for both chart-created and pre-existing SCC configurations.

Live OpenShift validation was not run.

## Stack

- `main` 
  - \#515
    - **feat(dataplane): add OpenShift compatibility support** :point\_left:

### DIFF
--- a/charts/dataplane/templates/serving/kourier-gateway-scc.yaml
+++ b/charts/dataplane/templates/serving/kourier-gateway-scc.yaml
@@ -1,0 +1,40 @@
+{{- if and (include "apps.enabled" .) (not .Values.zero_trust.enabled) .Values.serving.openShift.kourierGatewayScc.enabled .Values.serving.openShift.kourierGatewayScc.create }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ include "serving.kourierGateway.openShiftSccName" . }}
+  labels:
+    {{- include "serving.labels" . | nindent 4 }}
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities: []
+defaultAddCapabilities: []
+defaultAllowPrivilegeEscalation: false
+forbiddenSysctls:
+  - "*"
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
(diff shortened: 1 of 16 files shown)


charts/dataplane/values.openshift.yaml
```diff
@@ -0,0 +1,48 @@
+# OpenShift compatibility defaults.
```
mhotan, line 1: 👍 for creating an openshift specific overlay.

charts/dataplane/templates/serving/kourier-gateway-scc.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if and (include "apps.enabled" .) (not .Values.zero_trust.enabled) .Values.serving.openShift.kourierGatewayScc.enabled .Values.serving.openShift.kourierGatewayScc.create }}
```
mhotan, line 1: We should also gate on `if .Capabilities.APIVersions.Has "security.openshift.io/v1"`. This applies to all these openshift specific components.

mhotan:
@xjerod I am still debating how I feel about adding these Openshift specific manifests baked into the helm-chart templates.

Have we considered just guiding our customers by creating an https://github.com/unionai/helm-charts/tree/main/charts/dataplane/examples that adds `extraObjects` with these new objects?

Again not sure how I feel about this just yet

fiedlerNr9 (approved):

mhotan (approved):

PR #520, opened 2026-08-08
dataplane: default app serving to the vendored Knative Serving gateway

mhotan:
## Overview

Make the dataplane chart's **vendored** Knative Serving + Kourier + Envoy gateway (`templates/gateway/*`) the default app-serving delivery, decoupled from `zero_trust`, and disable the `knative-operator` subchart by default. Previously the vendored gateway rendered only under `zero_trust.enabled`, and app serving otherwise ran through the `knative-operator` subchart.

## What changed

- **`gateway.enabled`** (new, default `true`) selects vendored delivery; **`knative-operator.enabled`** default flipped to `false`.
- New helpers in `gateway/_helpers.tpl`:
  - `serving.useVendoredGateway` = `apps.enabled && gateway.enabled` — gates the Knative control-plane manifests (activator, autoscaler, controller, webhook, net-kourier-controller, configs, rbac, webhookconfigurations).
  - `serving.renderGateway` = `gateway.enabled && (apps.enabled || zero_trust.enabled)` — gates the Envoy plumbing (Envoy Deployment/Service + bootstrap), which fronts app serving **or** the zero-trust dataproxy routes.
  - `serving.useKnativeOperator` = `apps.enabled && !gateway.enabled` — gates the legacy `KnativeServing` CR path (escape hatch).
  - `serving.validateGateway` — fails the render when `zero_trust.enabled` is set without `gateway.enabled`.
- Operator app-ingress origin points at the gateway on the vendored path (instead of `kourier-internal`).

Zero trust is unchanged in behavior: it continues to layer its dataproxy + dataplane-ingress routes on the same Envoy gateway (now via `gateway.enabled`, which zero trust requires).

## Migration

Existing dataplanes running the knative-operator path must run the `knative-migration` Job before upgrading (it tears down the operator-managed `KnativeServing` and reconciles the Knative Serving CRDs). Fresh installs are unaffected.

## Testing

`helm template` verified across all gate paths:

- default → vendored gateway only; no `KnativeServing` CR; no knative-operator; zero-trust-only dataproxy backend absent.
- `gateway.enabled=false` → `KnativeServing` CR renders (operator path).
- `zero_trust.enabled` + `gateway.enabled=false` → render fails with the guard.

Snapshot regeneration to follow.

- `main` 
  - **dataplane: default app serving to the vendored Knative Serving gateway** :point\_left:
    - \#522


charts/dataplane/templates/gateway/service-public.yaml
```diff
@@ -0,0 +1,38 @@
...
+{{/*
+Public internet-facing LoadBalancer in front of the gateway Envoy, for
+separate-cluster dataplanes that serve apps but have no shared control-plane
+ingress LB to borrow. DNS `*.apps.<dp>.<domain>` (published by the per-DP public
+external-dns) resolves to this LB; Envoy terminates TLS on 8444 with the
+letsencrypt wildcard cert (443) and serves plain http on 8081 (80, for
+ACME/redirect). Intracluster dataplanes route apps through the shared CP
+ingress-nginx (ingress.serving.class) and leave this disabled.
+
+Annotations (LB scheme + external-dns hostname) come from the union_extension.
```
mhotan, line 11: Don't include union_extension reference. That is an internal union name.

charts/dataplane/Chart.yaml
```diff
@@ -27,6 +27,11 @@ dependencies:
   version: 2026.6.0
   alias: knative-operator
   condition: knative-operator.enabled
+# NOTE: the knative-operator -> vendored-gateway migration is NOT a subchart. It
```
mhotan: Remove the unnecessary comment

charts/dataplane/templates/gateway/service-public.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if and (include "apps.enabled" .) .Values.gateway.enabled .Values.gateway.publicLoadBalancer.enabled }}
+{{/*
```
mhotan, line 2: We need to justify why we need to add this and not leverage the dataplane extraObjects. Is this a generic way for tunneless public access to Apps?

charts/dataplane/templates/gateway/service-public.yaml
mhotan: Done — genericized to `gateway.publicLoadBalancer.annotations`; the internal `union_extension` name is gone (commit 3a4f2887).

charts/dataplane/Chart.yaml
mhotan: Done — removed the comment. The dependency block (`condition: knative-operator.enabled`) plus `knative-operator.enabled: false` in values already convey it, and RELEASE.md now documents the migration (3a4f2887).

charts/dataplane/templates/gateway/service-public.yaml
mhotan: Added justification to the template header (3a4f2887): yes, this is the generic tunnelless public entrypoint for app serving on a separate-cluster data plane. It's a first-class template rather than dataplane `extraObjects` because it's a reusable serving-gateway capability — toggled by one value (`gateway.publicLoadBalancer.enabled`) and sharing the gateway selector/labels — not per-environment glue. The only per-env part (LB scheme + external-dns hostname) stays in `gateway.publicLoadBalancer.annotations`.

charts/controlplane/values.yaml
```diff
@@ -2161,7 +2161,14 @@ flyte:
...
             sameSitePolicy: "LaxMode"
-            domain: ""
+            # Scope the session cookie to the control-plane host so it is sent to
+            # every subdomain, including per-dataplane app hostnames
+            # (*.apps.<cluster>.<host>). Without a Domain the cookie is host-only:
+            # a user logged into the console who opens an app is bounced to /login
+            # because the cookie never reaches the app subdomain. UNION_HOST is the
+            # public control-plane host (same value used for authorizedUris); the
+            # flyte-core subchart tpl-renders this against the shared global.
+            domain: '{{ .Values.global.UNION_HOST }}'
```
EngHabu, line 2171: Do we automatically add "*." as a prefix? don't we need that to allow cookies across "sub" domains of a domain?

EngHabu (approved):

mhotan:
Consolidated into #563 (`mike/serving-selfhosted-publicurl`), now the single self-hosted serving branch on latest main (vendored Knative gateway + NAP-compatible Spot routing + publicURLPattern). Branch retained.

PR #522, opened 2026-08-09
controlplane: drop the global apps publicURLPattern default (selfhosted)

mhotan:
An app's `public_url` isn't known until it's leased to a dataplane, and the dataplane self-declares its own apps domain via `connection_config` (`apps.domain`, DP name baked in); the control plane composes `{scheme}://{sub}.{domain}` at lease time. Empty the global `executions.apps.publicURLPattern` so app CREATE leaves `public_url` empty (filled at lease) instead of rendering a premature — and, with one `%s`, malformed — URL that failed IDL validation. Regenerate controlplane snapshots.

Stacked on the vendored-gateway PR (#520).

- `main` 
  - \#520
    - **controlplane: drop the global apps publicURLPattern default (selfhosted)** :point\_left:


charts/dataplane/templates/gateway/bootstrap-configmap.yaml
```diff
@@ -197,6 +241,23 @@ data:
                     interval: 30s
                     timeout: 5s
         {{- end }}
+        {{- if .Values.gateway.publicLoadBalancer.enabled }}
```
mhotan, line 244: We need to reconcile if this is really needed or we can repurpose 

```
{{- if include "apps.enabled" . }}
        - name: xds_cluster
          connect_timeout: 1s
          type: STRICT_DNS
          load_assignment:
            cluster_name: xds_cluster
            endpoints:
              lb_endpoints:
                - endpoint:
                    address:
                      socket_address:
                        address: net-kourier-controller
                        port_value: 18000
          typed_extension_protocol_options:
            envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
              "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
              explicit_http_config:
                http2_protocol_options:
                  connection_keepalive:
                    interval: 30s
                    timeout: 5s
        {{- end }}
```

mhotan: These are two different clusters that can't be reconciled:

- `xds_cluster` (net-kourier-controller:18000) is the **control channel** — Envoy's ADS/xDS connection to the Kourier controller, referenced by `dynamic_resources.ads_config`. It delivers the dynamic listener/cluster snapshot. It only exists when `apps.enabled` (no net-kourier-controller otherwise).
- `kourier_internal` (kourier-internal:80) is a **data-path upstream** — the static `apps_https:8443` listener terminates edge TLS and forwards decrypted app traffic here, and the internal Kourier Service host-routes it to the app's Knative revision.

One is HTTP/2 to the xDS management server; the other is an HTTP upstream to a different Service/port. The static listener can't target an xDS-managed cluster by name either — Kourier's generated cluster names aren't stable to reference from static config, which is exactly why we add a stable STRICT_DNS cluster to `kourier-internal:80`. So the block is needed as-is.

charts/controlplane/templates/common/_grpcroute-protected.yaml
```diff
@@ -52,6 +52,7 @@ spec:
     # cluster
     - matches:
         - method: {service: "cloudidl.cluster.ClusterService"}
+        - method: {service: "cloudidl.cluster.ClusterStatusService"}
```
mhotan, line 55: This is an unrelated fix.

charts/dataplane/values.yaml
```diff
@@ -2598,22 +2609,35 @@ gateway:
   # dataplanes (not co-located with the control plane) have no shared CP ingress
   # LB to borrow, so they expose the gateway Envoy directly: DNS
   # `*.apps.<dp>.<domain>` -> this LB -> Envoy, with TLS terminated at Envoy
-  # (8444) using the letsencrypt wildcard cert. Off by default; the terraform
-  # union_extension enables + annotates it (external-dns hostname, internet-facing
-  # scheme) for separate-cluster DPs. Intracluster apps route through the shared
-  # CP ingress-nginx instead (`ingress.serving.class`), so this stays off there.
+  # (8444) using the letsencrypt wildcard cert. Off by default; enable and
+  # annotate it (external-dns hostname, internet-facing scheme) for
```
mhotan, line 2649: Is this too specific to external-dns and letsencrypt?

charts/controlplane/values.yaml
```diff
@@ -1057,8 +1057,14 @@ services:
           # When true and no identity service is reachable, all runs display "Unknown"
           # even when created_by is populated in the DB.
           enrichIdentities: false
-          # App serving URL pattern (work in progress for self-hosted)
-          publicURLPattern: 'https://%s.apps.{{ .Values.global.UNION_HOST }}'
+          # Global fallback app-serving URL pattern (BYOC-style, two %s:
```
mhotan, line 1072: We need to update this comment. Given the internal changes we are making. 

This URL pattern is intended to be able to provide a stable URL for apps as apps are paused and restarted and potentially moved to different dataplane clusters in the same cluster pool.

However, as of today, setting this variable is not compatible with multi-dataplane.

mhotan: Updating as discussed — refining the flow this pattern assumes and why the global pattern is single-dataplane only.

**What it's for (BYOC): a stable URL across pause / restart / move.**
The pattern mints one hostname shape — `*.apps.<host>` — for every app, composed when the app is *created*, before it's placed on a dataplane. In BYOC a managed edge fronts that single wildcard hostname and forwards each request to whichever cluster currently runs the app, so an app can be paused or restarted and keep the exact same URL. That stability is the intent of a global pattern.

Caveat worth stating plainly: **BYOC multi-dataplane app serving does not work at this time either.** An app can only be deployed on a single cluster — spreading one app across dataplanes is an unimplemented placeholder. So the stable-URL benefit currently covers pause/restart on a single cluster, not multi-dataplane.

**Why one global pattern can't do multi-dataplane.**
A single wildcard hostname maps to one ingress target at a time — one DNS record, one target. If two dataplanes both claimed `*.apps.<host>`, it'd be last-writer-wins: whichever registered most recently owns the name, and the other is unreachable. The pattern also has no dataplane dimension (subdomain + org only) and is composed before a dataplane is chosen, so nothing distinguishes one dataplane from another.

**Why self-hosted empties it — and doesn't assume an edge.**
Self-hosted deliberately does **not** assume a managed edge. Instead each dataplane self-declares its own apps domain — dataplane name baked in — via `connection_config` (`apps.domain`), and the control plane composes each app's `public_url` at **lease** time from the dataplane that actually runs it (e.g. `…apps.dp-1.<domain>` vs `…apps.dp-2.<domain>`). Distinct per-dataplane hostnames, no collision — this is what makes multi-dataplane app serving work on self-hosted. Leaving the global pattern set would instead render a premature URL at create time (and, with a single `%s`, a malformed one that fails IDL validation).

**Trade-off:** a per-dataplane URL isn't stable if an app moves dataplanes, since the hostname carries the dataplane name. A stable cross-dataplane URL would need an edge/gateway that fans one name out to the right dataplane — future work, and something self-hosted intentionally doesn't assume today.

charts/dataplane/templates/gateway/envoy.yaml
```diff
@@ -193,6 +198,13 @@ spec:
         - name: config-volume
           configMap:
             name: {{ include "gateway.fullname" . }}-envoy-bootstrap
+        {{- if .Values.gateway.publicLoadBalancer.enabled }}
+        # Apps wildcard TLS cert (issued by the DP letsencrypt ClusterIssuer),
+        # served by the apps_https:8443 static listener in the bootstrap.
+        - name: apps-tls
+          secret:
+            secretName: dataplane-apps-letsencrypt-tls
```
katrogan, line 206: how is this created? is it documented?

charts/controlplane/values.yaml
```diff
@@ -1069,8 +1069,14 @@ services:
...
+          # subdomain, org). Empty on selfhosted: an app's public_url is not known
+          # until it is leased to a dataplane, and the dataplane self-declares its
+          # own apps domain via connection_config (apps.domain, dataplane name
+          # baked in). The control plane builds the URL as {scheme}://{sub}.{domain}
+          # at lease time. Leaving this set here would make app CREATE render a
+          # premature (and, with one %s, malformed) URL that fails IDL validation.
+          publicURLPattern: ''
```
katrogan, line 1079: is this change safe to make for our existing self-hosted customers? we'll stop populating PublicUrl when it's empty in apps service

charts/dataplane/templates/gateway/bootstrap-configmap.yaml
katrogan, line 244: do you need to also check if `apps.enabled`?

charts/dataplane/templates/gateway/envoy.yaml
```diff
@@ -175,6 +175,11 @@ spec:
           volumeMounts:
             - name: config-volume
               mountPath: /tmp/config
+            {{- if .Values.gateway.publicLoadBalancer.enabled }}
```
katrogan, line 178: do you need to also check if apps.enabled?

charts/dataplane/values.yaml
```diff
@@ -1658,9 +1667,9 @@ ingress:
...
     # nginx. Set false on separate-cluster dataplanes, which have no shared nginx
-    # and instead expose the gateway Envoy via gateway.publicLoadBalancer — the
-    # union_extension disables it there so the `<dp>-apps` Ingress isn't left
-    # dangling with no controller.
+    # and instead expose the gateway Envoy via gateway.publicLoadBalancer —
+    # disable it there so the `<dp>-apps` Ingress isn't left dangling with no
+    # controller.
     enabled: true
     # -- (Optional) Host override for serving ingress rule. Defaults to *.apps.<dataplane.cp.host>.
     hostOverride: '*.apps.{{ include "dataplane.cp.host" . }}'
```
katrogan, line 1675: is this right? shouldn't it be `apps.<CLUSTER>.<CP_HOST>`?

charts/dataplane/templates/gateway/bootstrap-configmap.yaml
```diff
@@ -162,6 +162,50 @@ data:
...
+        # Separate-cluster apps HTTPS. A static TLS listener that terminates the apps
+        # wildcard cert (letsencrypt, mounted at /etc/apps-tls) and forwards to the
+        # internal Kourier service. The public LoadBalancer targets 8443 here. The
+        # dynamic 8444 Kourier listener can't serve TLS without a per-app KIngress
+        # cert (net-certmanager), so this gives Envoy the wildcard directly.
+        - name: apps_https
+          address:
+            socket_address:
+              address: 0.0.0.0
+              port_value: 8443
```
katrogan, line 175: do we need to expose this port in envoy.yaml?

katrogan (approved):
I asked Claude and it suggested this improvement:

 The path a request actually takes                                                                                                                                       
                                                                                                                                                                          
  client → public LB :443                                                                                                                                                 
         → gateway pod A, Envoy :8443   (apps_https, TLS termination)                                                                                                     
         → DNS kourier-internal → ClusterIP VIP                                                                                                                           
         → kube-proxy DNAT → gateway pod ? :8081   (Kourier xDS listener)                                                                                                 
         → the Knative app

Hop 2 and hop 3 are the same set of processes. The request leaves pod A, goes out to the cluster VIP, and comes back into a gateway pod — possibly A, possibly B, chosen per-connection by kube-proxy. This adds a full extra hop, which affects latency and adds more stats/log noise churn

mhotan:
Folded into #563 (the consolidated self-hosted serving branch on latest main). Branch retained.

